PR #19, opened 2016-09-19
Bump mysql connector to 6.x

### LINKED ISSUE #18: Bump mysql-connector to 6.x
Can you bump the sql-connector to the latest version? We were happy with your library because it allowed us to get away from this approach here (http://labs.unacast.com/2016/02/18/cloud-sql-on-google-container-engine/), but as we are now bumping our dependencies it fails and we seem forced to revert back, increasing build time and container complexity / responsibilities. 

The error is ```java.lang.NoClassDefFoundError: com/mysql/jdbc/SocketFactory```. I believe it is pretty simple - the functionality is still present in 6.X, it just put in another package: ```com.mysql.cj.api.io.SocketFactory```

On a side note: Is there any rationale why the ```com.google.cloud.sql.mysql.SslSocketFactory``` is package local? It would be a lot easier for me to just rewrite this by creating my own SocketFactory without having to copy and duplicate your SSL and authentication stuff.

MartinSahlen:
Bump mysql java connector. All tests green. Fixes #18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*.iml
+.idea
+target/
--- a/pom.xml
+++ b/pom.xml
@@ -162,7 +162,7 @@
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <version>5.1.38</version>
+            <version>6.0.4</version>
         </dependency>
 
         <dependency>
--- a/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -29,13 +29,13 @@ import java.util.logging.Logger;
  *
  * <p>The heavy lifting is done by the singleton {@link SslSocketFactory} class.
  */
-public class SocketFactory implements com.mysql.jdbc.SocketFactory {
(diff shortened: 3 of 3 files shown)


laixer:
Thanks for the pull request. I would like to avoid implicitly upgrading all dependents on this package to 6.x especially since it's a development release. I would be OK publishing a separate artifact that works with the 6.x release, would that work for you? Let me know if you'd like to do it or want me to take care of it.

MartinSahlen:
Yes, I actually was not aware of the fact that 6.x is a development release until we had an internal discussion at my company after doing dependency bumping. We actually decided to revert back to 5.1.39 which is the latest stable release. 

As it is a pretty small job, I anyway think it would be cool to have this available for 6 if you can take care of it. Speaking against it is the fact that the mysql team can make incompatible changes that don't follow semantic versioning, but again, it would be cool to test it. I know it is working, as I got it working by just copying the code from this repo in its entirety and creating the SocketFactory that is in this pull request.

laixer:
New artifact for Connector/J 6.x is available:
http://search.maven.org/#artifactdetails%7Ccom.google.cloud.sql%7Cmysql-socket-factory-connector-j-6%7C1.0.2%7Cjar

PR #20, opened 2016-10-03
Add support for Connector/J 6.x

### LINKED ISSUE #18: Bump mysql-connector to 6.x

laixer:
Connector/J 6.x has a different location for the SocketFactory interface. 
This was done by splitting the project into multiple modules:
  core - mysql driver agnostic library code
  connector-j-5 - module for the 5.x version of Connector/J
  connector-j-6 - module for the 6.x version of Connector/J

The connector-j-6 module publishes a new artifact with id
'mysql-socket-factory-connector-j-6'.

Fixes #18.


connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.mysql;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.io.IOException;
+import java.net.Socket;
+import java.net.SocketException;
```
hfwang: nit: unused import

hfwang:
er, lack of proofreading: "it looks like keeping the split in the artifact id makes more sense for us anyway"

connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -0,0 +1,61 @@
...
+    String instanceName = props.getProperty("cloudSqlInstance");
+    checkArgument(
+            instanceName != null,
+            "cloudSqlInstance property not set. Please specify this property in the JDBC URL or "
+                    + "the connection Properties with value in form \"project:region:instance\"");
+
+    logger.info(String.format("Connecting to Cloud SQL instance [%s].", instanceName));
+
+    this.socket = SslSocketFactory.getInstance().create(instanceName);
+    return socket;
```
laixer: Done.

connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
laixer: Done.

hfwang:
LGTM

PR #21, opened 2016-10-03
Add some documentation for credentials.

laixer:


README.md
```diff
@@ -42,3 +42,10 @@ For example, if the instance connection name is `foo:bar:baz`, the JDBC URL
 would be 
 `jdbc:mysql://google/mydb?cloudSqlInstance=foo:bar:baz&socketFactory=com.google.cloud.sql.mysql.SocketFactory`
 
+## Credentials
```
Carrotman42: Do we want to add docs about Kubernetes as well?

Specifically, k8s doesn't recommend using default credentials even when running on GCE. Instead, they suggest downloading a service account and setting the environment variable for controlling credentials.

hfwang: This comment sent me down a rabbit hole of trying to figure out what the right way to do default credentials on k8s is.

Do you have a link to some documentation on this?

laixer: Good point, I've opened #22 to add k8s information. I'll need some extra time to do that as I want to get it working myself first.

evmin: Also would be helpful if it was possible to specify client_email and private_key (as defined in JSON file) as the URL parameters. This would allow for easy app parameterisation and avoid dealing with the key files.

PR #25, opened 2016-10-24
Add tool to help construct the JDBC URL.

laixer:


examples/getting-started/src/main/java/com/google/cloud/sql/mysql/example/GettingStarted.java
```diff
@@ -0,0 +1,327 @@
...
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.sqladmin.SQLAdmin;
+import com.google.api.services.sqladmin.model.DatabaseInstance;
+import com.google.api.services.sqladmin.model.InstancesListResponse;
+import com.google.common.collect.ImmutableSet;
+
+import java.io.Console;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.sql.*;
```
hfwang: I'd prefer you import classes explicitly.

examples/getting-started/src/main/java/com/google/cloud/sql/mysql/example/GettingStarted.java
```diff
@@ -0,0 +1,327 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
```
hfwang: This requires Java 8, will that be a problem?

examples/getting-started/src/main/java/com/google/cloud/sql/mysql/example/GettingStarted.java
```diff
@@ -0,0 +1,327 @@
...
+    for (int i = 0; i < instances.size(); i++) {
+      DatabaseInstance databaseInstance = instances.get(i);
+      System.out.println(String.format("%d: %s (%s)", i + 1, databaseInstance.getName(),
+          databaseInstance.getConnectionName()));
+    }
+
+    String instanceConnectionName = "";
+    while (instanceConnectionName.isEmpty()) {
+      System.out.print("Please enter the number of the instance you want to use [1]: ");
+      String line = console.readLine();
```
hfwang: Might be worth extracting the "select from a list" logic into its own method. It'd only have two users though.

examples/getting-started/src/main/java/com/google/cloud/sql/mysql/example/GettingStarted.java
laixer: Done, wasn't intentional. #justintellijthings

examples/getting-started/src/main/java/com/google/cloud/sql/mysql/example/GettingStarted.java
laixer: Should be fine for aux tools.

PR #26, opened 2016-10-26
Compute engine example.

laixer:


hfwang:
LGTM

Might be nice if there was a readme in that dir just saying what the example does

laixer:
Done.

PR #34, opened 2017-06-22
Add Postgres support for the Socket Factory.

laixer:
The core code in SslSocketFactory is reused as is.

The main difference between the MySQL implementation and the Postgres
implementation is that the user has to specify the instance name via the
`socketFactoryArg` property instead of a custom property as the Postgres
socket factory does not have access to the params.

There needs to be some cleanup in the code to move the common code out
of the 'mysql' package, but I'll leave it to a separate CL to keep this
small.

We also need to consider moving this code to a more 'generic' repo as
this one is mysql-specific, but that's a larger unertaking left to a
separate exercise.


hfwang:
LGTM

hfwang:
:shipit:

PR #36, opened 2017-06-27
Naming cleanup:

laixer:
- Move the 'core' classes from 'mysql' Java package to 'core' package.
 - Change 'mysql-socket-factory' to 'jdbc-socket-factory' in non-mysql
   specific artifact IDs.


hfwang:
LGTM

PR #47, opened 2017-08-17
Improve permission error message to mention curated role.

### LINKED ISSUE #48: Update permission error message to mention curated IAM role
pointed out by @joaoandremartins

laixer:
Fixes #48

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
@@ -469,7 +469,9 @@ public class SslSocketFactory {
             new RuntimeException(
                 String.format(
                     "Unable to obtain ephemeral certificate for Cloud SQL Instance [%s]. "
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -469,7 +469,9 @@ private X509Certificate obtainEphemeralCertificate(
             new RuntimeException(
                 String.format(
                     "Unable to obtain ephemeral certificate for Cloud SQL Instance [%s]. "
-                        + "Make sure %s Editor or Owner role on the project.",
+                        + "Ensure %s the sql.instances.connect permission "
```
hfwang: "Ensure %s *has* the..."

laixer: "has" is part of the substituted string because one alternative is "you have" instead of "foo has"

PR #52, opened 2017-09-29
Postgre on GAE Standard for Java8

lesv:
1. Update driver to support the GAE Postgre unix sockets
2. Add postgre on GAE for Java8 example
3. Update README
4. add .gitignore


lesv:
@laixer PTAL

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
+# Postgre SQL sample for Google App Engine for Java 8
```
iangudger: Postgres or PostgreSQL
https://en.wikipedia.org/wiki/PostgreSQL

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
...
+This sample demonstrates how to use [PostgreSql](https://cloud.google.com/sql/) on Google App
+Engine standard Java 8
+
+## Setup
+
+* If you haven't already, Download and initialize the [Cloud SDK](https://cloud.google.com/sdk/)
```
iangudger: Download -> download

Don't capitalize non-proper names in the middle of a sentence.

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
```diff
@@ -0,0 +1,122 @@
...
+    PrintWriter out = resp.getWriter();
+    resp.setContentType("text/plain");
+
+    // store only the first two octets of a users ip address
+    String userIp = req.getRemoteAddr();
+    InetAddress address = InetAddress.getByName(userIp);
+    if (address instanceof Inet6Address) {
+      // nest indexOf calls to find the second occurrence of a character in a string
+      // an alternative is to use Apache Commons Lang: StringUtils.ordinalIndexOf()
+      userIp = userIp.substring(0, userIp.indexOf(":", userIp.indexOf(":") + 1)) + ":*:*:*:*:*:*";
```
iangudger: IPv6 addresses can omit zero octets, and it appears that getRemoteAddr will do just that:
https://stackoverflow.com/questions/7589526/what-is-the-ip-format-returned-by-servletrequest-getremoteaddr

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
```diff
@@ -0,0 +1,122 @@
...
+
+    String path = req.getRequestURI();
+    if (path.startsWith("/favicon.ico")) {
+      return; // ignore the request for favicon.ico
+    }
+
+    PrintWriter out = resp.getWriter();
+    resp.setContentType("text/plain");
+
+    // store only the first two octets of a users ip address
```
iangudger: This seems unnecessary and overly complicates the code.

postgres/pom.xml
```diff
@@ -21,7 +22,17 @@
         <dependency>
             <groupId>com.google.cloud.sql</groupId>
             <artifactId>jdbc-socket-factory-core</artifactId>
-            <version>1.0.5-SNAPSHOT</version>
+            <version>1.0.4</version>
```
iangudger: Is there a reason why you aren't using the same version as above?

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Running locally
+When running locally, the SocketFactory can uses either the cloud_sql_proxy or it can get a certificate
+from the Cloud SQL and connect directly.
+
+### cloud_sql_proxy
+
+* `./cloud_sql_proxy -dir=/cloudsql &`
+* `export GAE_RUNTIME=java8`
+* follow the steps for Connecting directly below
```
iangudger: Capitalize follow?

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
...
+This sample demonstrates how to use [PostgreSql](https://cloud.google.com/sql/) on Google App
+Engine standard Java 8
+
+## Setup
+
+* If you haven't already, Download and initialize the [Cloud SDK](https://cloud.google.com/sdk/)
+
+    `gcloud init`
+
+* If you haven't already, Create an App Engine app within the current Google Cloud Project
```
iangudger: Consider adding a colon at the end of this (and similar) lines.

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+* [Create a Database](https://cloud.google.com/sql/docs/postgres/create-manage-databases)
+
+* [Create a user](https://cloud.google.com/sql/docs/postgres/create-manage-users)
+
+* Note the **Instance connection name** under Overview > properties it looks like 
+`Project:Region:Instance`
+
+## Running locally
+When running locally, the SocketFactory can uses either the cloud_sql_proxy or it can get a certificate
```
iangudger: uses -> use

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
...
+from the Cloud SQL and connect directly.
```
iangudger: Remove extra word the.

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
+# Postgre SQL sample for Google App Engine for Java 8
+This sample demonstrates how to use [PostgreSql](https://cloud.google.com/sql/) on Google App
+Engine standard Java 8
```
iangudger: Missing period at the end of sentence.

examples/postgres/appengine-standard-java8/README.md
```diff
@@ -0,0 +1,54 @@
...
+    `gcloud auth application-default login`
+
+
+* [Create an instance](https://cloud.google.com/sql/docs/postgres/create-instance)
+
+* [Create a Database](https://cloud.google.com/sql/docs/postgres/create-manage-databases)
+
+* [Create a user](https://cloud.google.com/sql/docs/postgres/create-manage-users)
+
+* Note the **Instance connection name** under Overview > properties it looks like
```
iangudger: Remove extra word it.

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
```diff
@@ -0,0 +1,122 @@
...
+public class PostgreSqlServlet extends HttpServlet {
+
+  Connection conn;
+
+  @Override
+  public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException,
+      ServletException {
+
+    final String createTableSql = "CREATE TABLE IF NOT EXISTS visits ( visit_id SERIAL NOT NULL, "
+        + "user_ip VARCHAR(46) NOT NULL, ts timestamp NOT NULL, "
```
iangudger: There is a Postgres type for storing IP addresses:
https://www.postgresql.org/docs/current/static/datatype-net-types.html

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
```diff
@@ -0,0 +1,122 @@
...
+      try {
+        Class.forName("org.postgresql.Driver");
+        conn = DriverManager.getConnection(url);
+      } catch (ClassNotFoundException e) {
+        throw new ServletException("Error loading JDBC Driver", e);
+      } catch (SQLException e) {
+        throw new ServletException("Unable to connect to PostGre", e);
+      }
+
+    } finally {
```
iangudger: Does an empty finally block serve a purpose?

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
lesv: IP addresses are PII in this instance.

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
lesv: IP addresses are PII in this instance.

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
lesv: This is the standard DevRel adopted for this stuff, (it's a standard for all languages), normally, I'd agree with you.

iangudger: You explicitly support running locally, but if IPv6 is used locally, I am pretty sure that this code will break. Localhost is ::1 in IPv6.

postgres/pom.xml
lesv: Procedure is:
1. LGTM & merge to this PR
2. Update all versions to production push new PR, merge.
3. Push new artifact to maven central
4. update all code to SNAPSHOT again, except the sample., Push a new PR, merge
5. done.  Update java-docs-samples to a version of this sample.
6. Update the docs.  Announce it.

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
iangudger: So? Collecting full IP addresses is standard operating procedure. Apache and Nginx for example do it by default.

postgres/pom.xml
iangudger: In some places `1.0.4` is used and other places `1.0.5-SNAPSHOT` is used. Shouldn't you be consistent?

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
iangudger: The guide for installing Apache doesn't suggest turning off logging IP addresses:
https://cloud.google.com/compute/docs/tutorials/basic-webserver-apache

lesv:
PTAL

postgres/pom.xml
lesv: 1.0.5-SNAPSHOT would be better, it just wasn't showing up for some reason.  I'll fix it.

lesv:
PTAL @iangudger

examples/postgres/appengine-standard-java8/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
lesv: Apache doesn't display it.

iangudger: But they log it, so storing it here is no different.

iangudger: When you retrieve the IP address from the database, you could get the binary representation with getAddress() and then display some of the bytes. That would be much simpler and avoid the correctness issues.
https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html

lesv:
I like your recent suggestion - will do this Wednesday so I can be a bit more detailed - as I was clearly too rushed last week.

lesv:
I've changed the PR to just the driver as it really should get pushed to maven-central

lesv:
@iangudger PTAL

postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java
```diff
@@ -32,6 +35,9 @@
  */
 public class SocketFactory extends javax.net.SocketFactory {
   private static final Logger logger = Logger.getLogger(SocketFactory.class.getName());
+  private static final String GaeRuntime = "GAE_RUNTIME";
```
iangudger: Since this constant is only used in one place and is named the same as its contents, is it needed? Wouldn't it be clearer to just use the string literal?

postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java
```diff
@@ -47,7 +53,15 @@ public SocketFactory(String instanceName) {
   @Override
   public Socket createSocket() throws IOException {
     logger.info(String.format("Connecting to Cloud SQL instance [%s].", instanceName));
-    return SslSocketFactory.getInstance().create(instanceName);
+    String runtime = System.getenv(GaeRuntime);
+
+    if (runtime == null || runtime.isEmpty()) {  // we aren't on Standard (not set local)
```
iangudger: What do you mean by ` (not set local)`? I understand that you are saying that the environmental variable with be set if and only if we are on standard, but I am confused by the parenthetical after that.

postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java
```diff
@@ -19,10 +19,13 @@
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.cloud.sql.core.SslSocketFactory;
+import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.util.logging.Logger;
+import org.newsclub.net.unix.AFUNIXSocket;
```
iangudger: Since we are using the same library that I suggested, why not get it from the official source like I did in the sample I set you rather than getting it from some unknown source?

I understand that you want something which is being maintained, but here are a couple of reasons why this might not be better:
* The original version works just fine. I have reviewed the code and it looks solid. This sort of thing is well understood and does not need frequent changes.
* Maintenance is only useful if the person doing the maintenance knows what they are doing. The original author knew what they were doing. Each change to something risks breaking it, so it is important to only make changes to address deficiencies rather than making changes for the sake of making changes.
* It is important for people who have write access on dependencies of official libraries to be reasonably trusted. They could make an intentionally malicious change. It might be worth vendoring an audited copy, for example, the original.

PR #67, opened 2018-04-19
Adds usage tracking

joaoandremartins:
Relies on the presence of a `USER_TOKEN_PROPERTY_NAME` property containing the user agent token.

An example of the client side usage is in https://github.com/spring-cloud/spring-cloud-gcp/commit/bc34b667a3595d2387a4057bf1a0d6f1b117e9d1

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
@@ -35,6 +35,14 @@ import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.util.concurrent.RateLimiter;
 
+import javax.annotation.Nullable;
+import javax.net.ssl.KeyManagerFactory;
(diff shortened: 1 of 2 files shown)


joaoandremartins:
@jabubake @lesv Could you please take a look and let me know what you think?

lesv:
And if not OpenCensus, we should grab the GAE / GCP request tokens

joaoandremartins:
@lesv I've never heard of OpenCensus. We're doing something similar to this with all other APIs we integrate with.

Can you point me in the right way so I can take a better look?

lesv:
https://opencensus.io/  is open census.  It's only a part of a possible solution.  Care to add Jabubake and I to a gvc to chat for 1/2 hr some time?

lesv:
and @kurtisvg

meltsufin:
@lesv I believe OpenCensus is designed more for application level request tracking, not for passing around `User-Agent` metadata.

PR #78, opened 2018-06-08
Add support for specifying a list of preferred IP types

hfwang:
Cloud SQL instances can have more IPs than just a primary IP, this change allows users to specify what IP type they would prefer to use when connecting.

### DIFF
--- a/connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -22,7 +22,6 @@ import com.google.cloud.sql.core.SslSocketFactory;
 import java.io.File;
 import java.io.IOException;
 import java.net.Socket;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
 import java.util.logging.Logger;
@@ -65,10 +64,11 @@ public class SocketFactory implements com.mysql.cj.api.io.SocketFactory {
       this.socket = UnixSocketChannel.open(socketAddress).socket();
     } else {
       // Default to SSL Socket
-      List<String> ipTypes =
-          listIpTypes(props.getProperty("cloudSqlIpTypes", "PUBLIC"));
       logger.info(String.format(
           "Connecting to Cloud SQL instance [%s] via ssl socket.", instanceName));
+      List<String> ipTypes =
+          SslSocketFactory.listIpTypes(
+              props.getProperty("cloudsqlIpTypes", SslSocketFactory.DEFAULT_IP_TYPES));
       this.socket = SslSocketFactory.getInstance().create(instanceName, ipTypes);
     }
     return this.socket;
@@ -84,18 +84,5 @@ public class SocketFactory implements com.mysql.cj.api.io.SocketFactory {
   @Override
   public Socket afterHandshake() {
     return socket;
-  }
-
-  private List<String> listIpTypes(String cloudSqlIpTypes) {
-    String[] rawTypes = cloudSqlIpTypes.split(",");
-    ArrayList<String> result = new ArrayList<>(rawTypes.length);
-    for (int i = 0; i < rawTypes.length; i++) {
-      if (rawTypes[i].trim().equalsIgnoreCase("PUBLIC")) {
-        result.add(i, "PRIMARY");
-      } else {
-        result.add(i, rawTypes[i].trim().toUpperCase());
-      }
-    }
-    return result;
   }
 }
--- a/connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -22,6 +22,8 @@ import com.google.cloud.sql.core.SslSocketFactory;
 import java.io.File;
 import java.io.IOException;
 import java.net.Socket;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
 import java.util.logging.Logger;
 import jnr.unixsocket.UnixSocketAddress;
@@ -63,16 +65,17 @@ public class SocketFactory implements com.mysql.cj.api.io.SocketFactory {
       this.socket = UnixSocketChannel.open(socketAddress).socket();
     } else {
       // Default to SSL Socket
+      List<String> ipTypes =
+          listIpTypes(props.getProperty("cloudSqlIpTypes", "PUBLIC"));
       logger.info(String.format(
           "Connecting to Cloud SQL instance [%s] via ssl socket.", instanceName));
-      this.socket = SslSocketFactory.getInstance().create(instanceName);
+      this.socket = SslSocketFactory.getInstance().create(instanceName, ipTypes);
     }
     return this.socket;
   }
 
   // Cloud SQL sockets always use TLS and the socket returned by connect above is already TLS-ready. It is fine
   // to implement these as no-ops.
-
   @Override
   public Socket beforeHandshake() {
     return socket;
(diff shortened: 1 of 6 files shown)


hfwang:
@kurtisvg, since we use the socketFactory and socketFactoryArgs, it's not obvious to me how to add an additional property to specify the ip type preference for postgres... for now, I've just fixed the code so it can compile, do you have any suggestions for how it could be implemented?

connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -63,6 +67,19 @@ public Socket connect(String host, int portNumber, Properties props, int loginTi
     return this.socket;
   }
 
+  private List<String> listIpTypes(String cloudSqlIpTypes) {
```
kurtisvg: Have you considered moving IpTypes into the SSLSocketFactory instead? Since it isn't needed until then, it would reduce the need to have this function in every class.

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -188,9 +190,26 @@ private static void logTestPropertyWarning(String property) {
...
   private SSLSocket createAndConfigureSocket(
-      String instanceName, CertificateCaching certificateCaching) throws IOException {
+      String instanceName,
+      List<String> ipTypes,
+      CertificateCaching certificateCaching)
+      throws IOException {
     InstanceSslInfo instanceSslInfo = getInstanceSslInfo(instanceName, certificateCaching);
-    String ipAddress = instanceSslInfo.getInstanceIpAddress();
+    String ipAddress = getPreferredIp(ipTypes, instanceSslInfo);
+    if (ipAddress == null) {
```
kurtisvg: Instead of doing an explicit check for null here, why not just throw the error inside the getPrefferedIp function?

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -615,21 +645,23 @@ public long getLastFailureMillis() {
...
         X509Certificate ephemeralCertificate,
         SSLSocketFactory sslSocketFactory) {
-      this.instanceIpAddress = instanceIpAddress;
       this.ephemeralCertificate = ephemeralCertificate;
       this.sslSocketFactory = sslSocketFactory;
     }
 
-    public String getInstanceIpAddress() {
-      return instanceIpAddress;
+    public void setInstanceIpAddress(String type, String ipAddress) {
```
kurtisvg: What is this supposed to be doing? I don't think set is the correct naming choice here, since it's really `put`ing the IP, which is sometimes adding a new object and sometimes replacing one. Can a type be replaced?

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -615,21 +645,23 @@ public long getLastFailureMillis() {
...
+      instanceIpAddresses.put(type == null ? "" : type.toUpperCase(), ipAddress);
+    }
+
+    public String getInstanceIpAddress(String type) {
+      return instanceIpAddresses.get(type.toUpperCase());
```
kurtisvg: Shouldn't type already be UpperCase by this point?

core/src/test/java/com/google/cloud/sql/core/SslSocketFactoryTest.java
```diff
@@ -170,20 +177,43 @@ public void create_throwsErrorForInvalidInstanceRegion() throws IOException {
...
-      sslSocketFactory.create("foo:bar:baz");
+      sslSocketFactory.create("foo:bar:baz", Arrays.asList("PRIMARY"));
       fail();
     } catch (IllegalArgumentException e) {
       assertThat(e.getMessage()).startsWith("Incorrect region value");
     }
   }
 
+  @Test
+  public void create_successfulPrivateConnection()
```
kurtisvg: Is there a specific reason for the underscore?

kurtisvg: Some comments in this test would be great

hfwang: The general style within Google for test naming is usually something like if there's a class Foo with a method bar, then in FooTest, if we want to test that bar works when baz is true, you would name the test method bar_bazIsTrue or something.

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
hfwang: I renamed it to put for now. There should be at most one ip per type, but I don't want the code to panic if that is ever not true.

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
hfwang: I think this is just meant to be defensive programming, so at least within the InstanceSslInfo class, the get and put treat the type as case insensitive, but that they do so by canonicalizing into upper case is an implementation detail.

core/src/test/java/com/google/cloud/sql/core/SslSocketFactoryTest.java
hfwang: Done

hfwang:
Hey Kurtis, I've taken another stab at this, and changed the postgres one to work via the new properties constructor.

postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java
```diff
@@ -41,8 +42,10 @@
   private static final String POSTGRES_SUFFIX = "/.s.PGSQL.5432";
 
   private static final String INSTANCE_PROPERTY_KEY = "cloudSqlInstance";
+  private static final String IP_TYPES_KEY = "cloudsqlIpTypes";
```
kurtisvg: Are you open to renaming this to just `ipTypes`? If not, can we at least use `cloudSqlIpTypes` to be remain consistent?

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -78,6 +81,7 @@
 public class SslSocketFactory {
   private static final Logger logger = Logger.getLogger(SslSocketFactory.class.getName());
 
+  public static final String DEFAULT_IP_TYPES = "PUBLIC,PRIVATE";
```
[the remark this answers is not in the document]
kurtisvg: "PRIMARY" is the API call, but "PUBLIC" is more understandable, so we will be using "PUBLIC". Proxy will be updated to match.

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -203,6 +211,50 @@ private SSLSocket createAndConfigureSocket(
...
+    String ipAddress = null;
+    for (String ipType : ipTypes) {
+      ipAddress = instanceSslInfo.getInstanceIpAddress(ipType);
+      if (ipAddress != null) {
+        break;
+      }
+    }
+
+    if (ipAddress == null) {
+      StringBuilder sb = new StringBuilder();
```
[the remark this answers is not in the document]
hfwang: done

postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java
hfwang: done, let's go with ipTypes.

core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -203,6 +211,50 @@ private SSLSocket createAndConfigureSocket(
     return sslSocket;
   }
 
+  /**
+   * Converts the string property of IP types to a list by splitting by commas, and upper-casing.
+   */
+  public static List<String> listIpTypes(String cloudSqlIpTypes) {
```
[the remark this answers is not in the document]
hfwang: I'm a little uncomfortable with doing validation or making it an enum because it makes it harder to make changes later (particularly for enums, if we introduce a new value or remove one, it would require a new build)

PR #79, opened 2018-06-08
Correctly sets user agent

joaoandremartins:
User agent set in the previous way was being overridden by the
API code.
This should now make it possible to correctly track usage, which
isn't happening at the moment.


core/src/main/java/com/google/cloud/sql/core/SslSocketFactory.java
```diff
@@ -141,13 +142,13 @@ public static synchronized SslSocketFactory getInstance() {
       }
 
       String userToken = System.getProperty(USER_TOKEN_PROPERTY_NAME);
-      Credential credential = credentialFactory.create();
+      if (userToken != null) {
+        applicationName = userToken;
```
meltsufin: Can you do this inline where you call `setApplicationName()`, and then avoid updating a static variable?

joaoandremartins:
@jabubake any ideas when we can have a new release here? :)

PR #90, opened 2018-07-09
Fix Postgres dependency version typo and make maven and gradle versio…

anuraaga:
…ns consistent.


README.md
```diff
@@ -84,7 +84,7 @@ When specifying the JDBC connection URL, add the additional parameters:
 | Property         | Value         |
 | ---------------- | ------------- |
 | socketFactory    | com.google.cloud.sql.postgres.SocketFactory |
-| socketFactoryArg | The instance connection name (found on the instance details) |
```
kurtisvg: This is actually correct as is - the Postgres JDBC driver doesn't have access to all the properties like the the mysql-connector-j driver does. See the [driver documentation](https://jdbc.postgresql.org/documentation/head/connect.html) and our [page](https://cloud.google.com/appengine/docs/standard/java/cloud-sql/using-cloud-sql-postgres) on Postgres JDBC urls.

anuraaga: Yeah but this section is for MySQL - the next one is for postgres which I left unchanged

kurtisvg: You're right, my mistake!

PR #97, opened 2018-08-06
Use built in base64 to allow use on jdk9+

marshallpierce:
`javax.xml.bind.DatatypeConverter` is not available out of the box on modularized JVMs, but since this project is already targeting 1.8, we can use `java.util.Base64` which was introduced then.

I also updated a few maven plugins so the build wouldn't die on current JDKs.


marshallpierce:
I signed it!

PR #103, opened 2018-09-25
Update README.md

sergei-ivanov:
Added firewall configuration notes. It appears that the socket factory attempts to connect to port 3307 on the Cloud SQL instance, but that is not documented anywhere and is not obvious at all. The port number is currently hard-coded as a constant: `com.google.cloud.sql.core.SslSocketFactory.DEFAULT_SERVER_PROXY_PORT` and it took me a while to find it. I hope that the proposed documentation will make troubleshooting easier for other people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,20 +123,7 @@ SocketFactory to connect with an instance's associated private IP. Default value
 
 ### Firewall configuration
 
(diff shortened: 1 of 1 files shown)


kurtisvg:
@sergei-ivanov Thanks for pointing this out - I agree it should be documented. I've tweaked the phrasing a little for brevity. Would you check it over to make sure I didn't leave anything out?

sergei-ivanov:
@kurtisvg LGTM, thanks for editing it!

PR #107, opened 2018-11-08
Support new SocketFactory API in mysql-connector-java 8.0.13.

anuraaga:
mysql-connector-java 8.0.13 introduced a major backwards incompatible change in the `SocketFactory` API. This makes sure the cloud SocketFactory implements both old and new APIs.

```
Exception in thread "main" java.lang.AbstractMethodError: Receiver class com.google.cloud.sql.mysql.SocketFactory does not define or inherit an implementation of the resolved method abstract connect(Ljava/lang/String;ILcom/mysql/cj/conf/PropertySet;I)Ljava/io/Closeable; of interface com.mysql.cj.protocol.SocketFactory.
	at com.mysql.cj.protocol.a.NativeSocketConnection.connect(NativeSocketConnection.java:65)
	at com.mysql.cj.NativeSession.connect(NativeSession.java:152)
	at com.mysql.cj.jdbc.ConnectionImpl.connectOneTryOnly(ConnectionImpl.java:955)
	at com.mysql.cj.jdbc.ConnectionImpl.createNewIO(ConnectionImpl.java:825)
	at com.mysql.cj.jdbc.ConnectionImpl.<init>(ConnectionImpl.java:455)
	at com.mysql.cj.jdbc.ConnectionImpl.getInstance(ConnectionImpl.java:240)
	at com.mysql.cj.jdbc.NonRegisteringDriver.connect(NonRegisteringDriver.java:207)
	at com.zaxxer.hikari.util.DriverDataSource.getConnection(DriverDataSource.java:136)
	at com.zaxxer.hikari.pool.PoolBase.newConnection(PoolBase.java:369)
	at com.zaxxer.hikari.pool.PoolBase.newPoolEntry(PoolBase.java:198)
	at com.zaxxer.hikari.pool.HikariPool.createPoolEntry(HikariPool.java:467)
	at com.zaxxer.hikari.pool.HikariPool.checkFailFast(HikariPool.java:541)
	at com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:115)

```

I verified the change on the three GA releases, 8.0.11, 8.0.12, and 8.0.13.

I ran into a strange behavior of Java, I believe, where previously it was fine to implicitly implement a method that returns `<T extends Closeable> T` with a method that returns `Socket`, but from what I can tell it only works when compiling a class that actually implements the interface (guess Java generates bytecode for the concrete and `Closeable` version when it sees such an interface override). Now that the method stopped having an interface method it was implementing, such a helper wasn't generated and I couldn't get the new code to run on `8.0.11` until changing the signature from `Socket` to `<T extends Closeable> T`. Since this is actually closer to the definition in `8.0.11`, it is a compatible change.

### DIFF
--- a/connector-j-8/pom.xml
+++ b/connector-j-8/pom.xml
@@ -21,7 +21,7 @@
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <version>8.0.11</version>
+            <version>8.0.13</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
--- a/connector-j-8/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-8/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -19,9 +19,11 @@ package com.google.cloud.sql.mysql;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.cloud.sql.core.SslSocketFactory;
+import com.mysql.cj.conf.PropertySet;
 import com.mysql.cj.protocol.ServerSession;
 import com.mysql.cj.protocol.SocketConnection;
 
+import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.net.Socket;
@@ -44,7 +46,14 @@ public class SocketFactory implements com.mysql.cj.protocol.SocketFactory {
   private Socket socket;
 
   @Override
-  public Socket connect(String host, int portNumber, Properties props, int loginTimeout) throws IOException {
+  public <T extends Closeable> T connect(String host, int portNumber, PropertySet props, int loginTimeout)
+      throws IOException {
+    return connect(host, portNumber, props.exposeAsProperties(), loginTimeout);
+  }
+
+  // This is the API before mysql-connector-java 8.0.13
+  public <T extends Closeable> T connect(String host, int portNumber, Properties props, int loginTimeout)
+      throws IOException {
     String instanceName = props.getProperty("cloudSqlInstance");
     checkArgument(
             instanceName != null,
@@ -74,7 +83,9 @@ public class SocketFactory implements com.mysql.cj.protocol.SocketFactory {
               props.getProperty("ipTypes", SslSocketFactory.DEFAULT_IP_TYPES));
       this.socket = SslSocketFactory.getInstance().create(instanceName, ipTypes);
     }
-    return this.socket;
+    @SuppressWarnings("unchecked")
+    T castSocket = (T) this.socket;
+    return castSocket;
   }
 
 


kurtisvg:
Hey @anuraaga, thanks for the PR. 

Could you present a link to the connector notes where this change is discussed? I wasn't able to find anything with a quick search.

Would you also expand on the `<T extends Closeable> T` vs `Socket` change? I'm not quite sure what you mean by "the method stopped having an interface method it was implementing, such a helper wasn't generated". Could you provide a little more explanation on what is happening and why this change is necessary?

anuraaga:
Oops sorry had meant to link to this commit.

https://github.com/mysql/mysql-connector-j/commit/3f44adb9bbf6d53528294e0668874724df8edd4e#diff-a8538d4116e1db5fe21f876c58fc1360

Unfortunately don't know of any other discussion besides it.

Also sorry for the poor explanation of `T extends Closeable` and `Socket` - TBH I'm not totally sure either. What I found was when compiling vs 8.0.13, even though the old method was completely unchanged, 8.0.12 could not find the method at runtime - changing the return type to be exactly the same as in 8.0.12, `T extends Closeable`, fixed it. I can only presume that javac generates code differently when a method "partially matches" an interface method compared to when the method is defined by itself, but it's behavior I've never heard of before.

lesv:
Surprised the folks at Oracle didn't mention this in the release notes.  It would be nice if they had, or at least mentioned it somewhere explicitly.

Make sure to test w/ Java 7, Java 8, Java 9, Java 10, and Java 11.  You can do a `mvn install` then reference the `-SNAPSHOT` in your sample pom.xml. 

I'm a bit concerned as PropertySet w/ introduced in Java 10.  But we'll need to adopt something like this PR before we support Java 11.

lesv:
[8.0.13](https://dev.mysql.com/doc/relnotes/connector-j/8.0/en/news-8-0-13.html) or [8.0.12](https://dev.mysql.com/doc/relnotes/connector-j/8.0/en/news-8-0-12.html) release notes don't mention it.

Though the 8.0.12 notes say:

**Name change of the com.mysql.jdbc.SocketFactory interface to com.mysql.cj.api.io.SocketFactory caused backward incompatibility for older Connector/J applications. The old interface has now been reimplemented to avoid the incompatibility. (Bug #25223137, Bug #84099)**

lesv:
[#84099](https://bugs.mysql.com/bug.php?id=84099) doesn't appear to be related.

anuraaga:
Oh gosh - thanks for the find. When getting ready to upgrade to mysql 8, I tried first with the RC and found it had a new `SocketFactory` API and implemented mysql-socket-factory-connector-j-8 to be ready for it. It didn't occur to me at all that with `8.0.11`, it might be possible to just use the old `SocketFactory`. I confirmed that all GA versions of mysql-connector-java, 8.0.11, 8.0.12, and 8.0.13 work with the vanilla mysql-socket-factory.

I'm sorry for unnecessarily forking the factory and causing maintenance headaches :( I have sent #108 as a proposal for cleaning up the README to reflect the situation.

kurtisvg:
@anuraaga - No need to apologize for forking. In fact, I think remove the other artifacts might be premature at this point. According to [#84099](https://bugs.mysql.com/bug.php?id=84099) it looks like the older interface wasn't implemented until `8.0.11`, so it might still be necessary for anyone using the older versions of the driver. 

I'll discuss this with a few people and follow up on #108 to see if we want to completely remove the newer version or possibly keep it until we move our target goal to 8.x+.

anuraaga:
Thanks @kurtisvg - I feel as if versions before 8.0.11, all of which were either developer preview or RC versions, shouldn't expect much longer maintenance but look forward to hearing what you all think :)

anuraaga:
@kurtisvg As discussed in #108 I went ahead and reponed this so the factory can support the API in 8.0.13

kurtisvg:
Thanks Les - this is for the j8 connector only, so I think that is the only version we need to be concerned about.

kurtisvg:
@anuraaga Sorry this took so long for me to get too. Thanks again - your work is very much appreciated.

PR #108, opened 2018-11-09
Remove connector-j-6/8 flavors of the factory as the vanilla version …

anuraaga:
…works with all GA versions of the connector.

During the conversation in #107 I found that the vanilla version of the factory can support all GA versions of mysql-connector-java. So it doesn't make sense to keep the `j-8` flavor of the connector. As GA 8.x has been around for a while, it also seems like a good time to remove `j-6` to make the docs clear.


kurtisvg:
Hey so to follow up on this, we are going to undergo a formal deprecation for the j5 and j6 versions of the library. While the j5 version of the library is still compatible with the j8 version, aiming for the most recent version of the interface could potentially be beneficial. 

So at some point in the near future, we will mark the j5 and j6 versions as deprecated, and after a 1 year period remove them from the library. As such, I'm going to go ahead and close this PR, but thanks for taking the time to surface this issue with this PR and #107 - it's much appreciated.

anuraaga:
Thanks - this makes sense. But does it mean I should reopen #107? Without that change, if j5 were deprecated there would be no non-deprecated version of the factory that works with 8.0.13

Issue #110, opened 2018-12-14
Update App Engine detection for new Java Standard GAE runtime 


ludoch:
We are working on a newer Java Standard GAE runtime, without the compat layer...
In this execution env, "com.google.appengine.runtime.environment" is not set at all. only GAE_ENV
So
https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/master/connector-j-8/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java#L57

does not work.
We are currently forced to use the undocumented CLOUD_SQL_FORCE_UNIX_SOCKET env var to make it work.

ludoch:
In this new GAE Standads Java env com.google.appengine.runtime.environment is neither set to true or false
In the existing GAE Java8 Standard env, it is either set to true or to false.

kurtisvg:
The `CLOUD_SQL_FORCE_UNIX_SOCKET` env var is documented in the project's [README](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/master/README.md#connection-via-unix-sockets).

Moving forward, what is the preferred method for detecting if the SF is executing in this new environment?

ludoch:
String gaeEnv = System.getenv("GAE_ENV");

is still "standard"

    String runEnv = System.getProperty("com.google.appengine.runtime.environment");

should be null in this environment (so neither prod nor dev).

kurtisvg:
Offline info + some summary on this:

We can detect the new environment when `GAE_ENV="standard"` and `GAE_RUNTIME="java11"`. 

We should be able to support both versions with the following logic: 
`GAE_ENV == "standard"` and either `com.google.appengine.runtime.environment == "Production"` or `GAE_RUNTIME == "java11"`

PR #116, opened 2019-01-18
Updated GAE check for Java11 compatibility.

kurtisvg:
Addressed #110.

I've confirmed this works on Java 8 GAE env - @ludoch can you verify it works for the new runtime?

### DIFF
--- a/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -35,6 +35,7 @@ import jnr.unixsocket.UnixSocketChannel;
  * <p>The heavy lifting is done by the singleton {@link SslSocketFactory} class.
  */
 public class SocketFactory implements com.mysql.jdbc.SocketFactory {
+
   private static final Logger logger = Logger.getLogger(SocketFactory.class.getName());
   private static final String CloudSqlPrefix = "/cloudsql/";
 
@@ -84,7 +85,7 @@ public class SocketFactory implements com.mysql.jdbc.SocketFactory {
 
   // TODO(kurtisvg) move this check into a shared class
   // Returns @{code true} if running in a Google App Engine Standard runtime.
-  private boolean runningOnGaeStandard(){
+  private boolean runningOnGaeStandard() {
     // gaeEnv="standard" indicates standard instances
     String gaeEnv = System.getenv("GAE_ENV");
     // runEnv="Production" requires to rule out Java 8 emulated environments
@@ -92,6 +93,7 @@ public class SocketFactory implements com.mysql.jdbc.SocketFactory {
     // gaeRuntime="java11" in Java 11 environments (no emulated environments)
     String gaeRuntime = System.getenv("GAE_RUNTIME");
 
-    return "standard".equals(gaeEnv) && ("Production".equals(runEnv) || "java11".equals(gaeRuntime));
+    return "standard".equals(gaeEnv) && ("Production".equals(runEnv) || "java11"
+        .equals(gaeRuntime));
   }
 }
--- a/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -54,15 +54,15 @@ public class SocketFactory implements com.mysql.jdbc.SocketFactory {
 
     // If running on GAE Standard, connect with unix socket
     if (forceUnixSocket || runningOnGaeStandard()) {
-      logger.info(String.format(
-          "Connecting to Cloud SQL instance [%s] via unix socket.", instanceName));
-      UnixSocketAddress socketAddress = new UnixSocketAddress(
-          new File(CloudSqlPrefix + instanceName));
+      logger.info(
+          String.format("Connecting to Cloud SQL instance [%s] via unix socket.", instanceName));
+      UnixSocketAddress socketAddress =
+          new UnixSocketAddress(new File(CloudSqlPrefix + instanceName));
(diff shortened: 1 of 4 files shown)


connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -85,4 +81,17 @@ public Socket beforeHandshake() {
   public Socket afterHandshake() {
     return socket;
   }
+
```
netdpb: Remove this extra blank line. Do you use google-java-format? It's a good idea.

connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -85,4 +81,17 @@ public Socket beforeHandshake() {
...
+
+  // Returns True if running in a Google App Engine Standard runtime, otherwise False.
```
netdpb:
```suggestion
  /** Returns {@code true} if running in a Google App Engine Standard runtime, otherwise {@code false}. */
```

I know it's private, but if you're going to document what the method does you might as well use Javadoc. Also, you could probably skip "otherwise false" since that's implied.

connector-j-8/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -102,4 +98,16 @@ public Socket performTlsHandshake(SocketConnection socketConnection, ServerSessi
 
   @Override
   public void afterHandshake() {}
+
+  // Returns True if running in a Google App Engine Standard runtime, otherwise False.
+  private boolean runningOnGaeStandard(){
```
netdpb: Is there anywhere common you could put this method, maybe as a static method, instead of copying it three times?

kurtisvg: The only place that I'm aware of would be in the core module. We could add a new class to hold helpers, or add it in the SslSocketFactory (but I'm not sure it makes sense there). What are your thoughts?

netdpb: Do these modules already depend on the core module? If so, that sounds like the right place. I think it makes sense to add it to a new class. I usually prefer class names that describe the class rather than "helper". This is about the runtime environment, right, so how about a class called `Environment`?

kurtisvg: Yes, all 4 depend on the core module. I think it would be best to handle this in a separate PR, and perhaps consider moving more of the guts of the connect methods to a shared class as well. I'll add it as an action item to the doc.

connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
kurtisvg: Done.

connector-j-6/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
kurtisvg: Yes, but there are a few other violations so I've avoided using it up until this point. I was planning on adding a check with the check-style plugin as well.

kurtisvg: I've updated the 4 SocketFactory files, but will push the update to core to a different PR.

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -83,15 +83,17 @@ public Socket afterHandshake() {
     return socket;
   }
 
-  // Returns True if running in a Google App Engine Standard runtime, otherwise False.
-  private boolean runningOnGaeStandard(){
+  // TODO(kurtisvg) move this check into a shared class
+  // Returns @{code true} if running in a Google App Engine Standard runtime.
```
netdpb: You need `/** ... */` for it to be Javadoc.

And for code snippets it's `{@code true}`.

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -84,14 +85,15 @@ public Socket afterHandshake() {
...
+  private boolean runningOnGaeStandard() {
     // gaeEnv="standard" indicates standard instances
     String gaeEnv = System.getenv("GAE_ENV");
     // runEnv="Production" requires to rule out Java 8 emulated environments
     String runEnv = System.getProperty("com.google.appengine.runtime.environment");
     // gaeRuntime="java11" in Java 11 environments (no emulated environments)
     String gaeRuntime = System.getenv("GAE_RUNTIME");
 
-    return "standard".equals(gaeEnv) && ("Production".equals(runEnv) || "java11".equals(gaeRuntime));
+    return "standard".equals(gaeEnv) && ("Production".equals(runEnv) || "java11"
```
netdpb: This doesn't look like it was formatted by [google-java-format](https://github.com/google/google-java-format). That's available as an IntelliJ plugin as well as a command-line tool.

Generally we apply it in the mode where it only affects lines that were already touched in a PR, in order to avoid extraneous changes, unless there's a reformat-only PR. But it's up to you.

kurtisvg: Ok, I believe I've run the correct one now. (I apparently only had it enabled for g3 files)

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
kurtisvg: My mistake, fixed.

PR #122, opened 2019-01-25
Enforce google-style-format and added checkstyle check for CI testing. 

kurtisvg:


connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -28,6 +28,7 @@
 import jnr.unixsocket.UnixSocketAddress;
 import jnr.unixsocket.UnixSocketChannel;
 
+
```
netdpb: Only one blank line needed here.

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -84,7 +85,9 @@ public Socket afterHandshake() {
     return socket;
   }
 
-  /** Returns {@code true} if running in a Google App Engine Standard runtime. */
+  /**
```
netdpb: Hmm. google-java-format usually prefers single-line Javadoc if it fits. Are you running that?

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
```diff
@@ -20,7 +20,7 @@
 
 /**
  * Factory for creating {@link Credential}s for interaction with Cloud SQL Admin API.
- */
+ * */
```
netdpb:
```suggestion
 */
```

google-java-format should have caught this.

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
netdpb: OK. I don't personally use Checkstyle, even with Google settings. What value do you find it has?

pom.xml
```diff
@@ -1,183 +1,216 @@
...
+    </snapshotRepository>
+
+    <repository>
+      <id>ossrh</id>
+      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+    </repository>
+  </distributionManagement>
+
+  <scm>
+    <connection>scm:git:https://github.com/GoogleCloudPlatform/cloud-sql-mysql-socket-factory.git
```
netdpb: This looks odd.

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
kurtisvg: Checkstyle is enforcement of the style. I spent some time experimenting with using google-java-format and diffing, but checkstyle provides a better explanation of why something is incorrect and is more accessible to Windows users.

PR #126, opened 2019-02-11
Consolidate "connect" logic from individual factories into core SocketFactory.

### LINKED ISSUE #113: 1.0.11 is not backwards compatible to 1.0.10 (postgresql)
in 1.0.10, you could include the `socketFactoryArg` parameter in your jdbc url in order to pass the cloudSqlInstance name. In 1.0.11, doing this leads to the value not being found by the SocketFactory constructor, failing the call to `checkArgument` in line 53 of `com.google.cloud.sql.postgres.SocketFactory`.

IMHO, API breakage should not happen on micro bumps. There also is no explicit reference to this in the release notes AFAICS.

kurtisvg:
Consolidates the "connect" function logic from being implemented in each Factory into the core Socket Factory. 

Notable changes:
* Renamed core "SslSocketFactory" to "CoreSocketFactory" to better reflect it now returns multiple types of sockets
* Renamed the "create" function to "createSslSocketFactory"
* Implemented a new "create" function containing the ssl/unix socket determination and implemented in each factory 
* As a side effect, cleaned up the way "SocketFactoryArg" deprecation was handled that fixes #113

### DIFF
--- a/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -66,7 +66,7 @@ public class SocketFactory implements com.mysql.jdbc.SocketFactory {
       List<String> ipTypes =
           CoreSocketFactory.listIpTypes(
               props.getProperty("ipTypes", CoreSocketFactory.DEFAULT_IP_TYPES));
-      this.socket = CoreSocketFactory.getInstance().create(instanceName, ipTypes);
+      this.socket = CoreSocketFactory.getInstance().createSslSocket(instanceName, ipTypes);
     }
     return this.socket;
   }
--- a/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -33,7 +33,8 @@ public class SocketFactory implements com.mysql.jdbc.SocketFactory {
 
   @Override
   public Socket connect(String hostname, int portNumber, Properties props) throws IOException {
-    socket = CoreSocketFactory.getInstance().connect(props);
+    socket =
+        CoreSocketFactory.getInstance().connect(props, CoreSocketFactory.MYSQL_SOCKET_FILE_FORMAT);
     return socket;
   }
 
--- a/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -18,7 +18,7 @@ package com.google.cloud.sql.mysql;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
-import com.google.cloud.sql.core.SslSocketFactory;
+import com.google.cloud.sql.core.CoreSocketFactory;
 import java.io.File;
 import java.io.IOException;
 import java.net.Socket;
@@ -32,7 +32,7 @@ import jnr.unixsocket.UnixSocketChannel;
  * A MySQL {@link SocketFactory} that establishes a secure connection to a Cloud SQL instance using
  * ephemeral certificates.
  *
- * <p>The heavy lifting is done by the singleton {@link SslSocketFactory} class.
+ * <p>The heavy lifting is done by the singleton {@link CoreSocketFactory} class.
  */
 public class SocketFactory implements com.mysql.jdbc.SocketFactory {
 
@@ -64,9 +64,9 @@ public class SocketFactory implements com.mysql.jdbc.SocketFactory {
       logger.info(
           String.format("Connecting to Cloud SQL instance [%s] via ssl socket.", instanceName));
       List<String> ipTypes =
-          SslSocketFactory.listIpTypes(
-              props.getProperty("ipTypes", SslSocketFactory.DEFAULT_IP_TYPES));
-      this.socket = SslSocketFactory.getInstance().create(instanceName, ipTypes);
+          CoreSocketFactory.listIpTypes(
+              props.getProperty("ipTypes", CoreSocketFactory.DEFAULT_IP_TYPES));
+      this.socket = CoreSocketFactory.getInstance().create(instanceName, ipTypes);
     }
     return this.socket;
   }
--- a/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
+++ b/connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
@@ -16,17 +16,10 @@
 
 package com.google.cloud.sql.mysql;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 import com.google.cloud.sql.core.CoreSocketFactory;
-import java.io.File;
 import java.io.IOException;
 import java.net.Socket;
-import java.util.List;
 import java.util.Properties;
-import java.util.logging.Logger;
-import jnr.unixsocket.UnixSocketAddress;
-import jnr.unixsocket.UnixSocketChannel;
 
 /**
  * A MySQL {@link SocketFactory} that establishes a secure connection to a Cloud SQL instance using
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -77,11 +82,15 @@
  *
  * <p>The API of this class is subject to change without notice.
  */
-public class SslSocketFactory {
-  private static final Logger logger = Logger.getLogger(SslSocketFactory.class.getName());
+public class CoreSocketFactory {
+  public static final String CLOUD_SQL_INSTANCE_PROPERTY = "cloudSqlInstance";
```
netdpb: Do these all need to be public?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -93,7 +102,7 @@
   private static final int DEFAULT_SERVER_PROXY_PORT = 3307;
   private static final int RSA_KEY_SIZE = 2048;
 
-  private static SslSocketFactory sslSocketFactory;
+  private static CoreSocketFactory sslSocketFactory;
```
netdpb: Can you rename the field to be consistent?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -126,11 +135,12 @@
   }
 
   /**
-   * Returns the SslSocketFactory singleton, which can be used to create SslSockets to Cloud SQL.
+   * Returns the CoreSocketFactory singleton, which can be used to createSslSocket SslSockets to
+   * Cloud SQL.
    *
-   * @return the SslSocketFactory singleton.
+   * @return the CoreSocketFactory singleton.
```
netdpb: You don't need this if it simply repeats what's in the summary.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -126,11 +135,12 @@
   }
 
   /**
-   * Returns the SslSocketFactory singleton, which can be used to create SslSockets to Cloud SQL.
+   * Returns the CoreSocketFactory singleton, which can be used to createSslSocket SslSockets to
```
netdpb:
```suggestion
   * Returns the {@link CoreSocketFactory} singleton.
```

What it does is documented elsewhere.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -152,12 +162,61 @@ public static synchronized SslSocketFactory getInstance() {
...
+          String.format(
+              "Connecting to Cloud SQL instance [%s] via unix socket.", csqlInstanceName));
+      UnixSocketAddress socketAddress =
+          new UnixSocketAddress(new File(String.format(socketPathFormat, csqlInstanceName)));
+      return UnixSocketChannel.open(socketAddress).socket();
+    }
+
+    logger.info(
+        String.format("Connecting to Cloud SQL instance [%s] via SSL socket.", csqlInstanceName));
+    return CoreSocketFactory.getInstance().createSslSocket(csqlInstanceName, ipTypes);
```
netdpb:
```suggestion
    return getInstance().createSslSocket(csqlInstanceName, ipTypes);
```

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -77,11 +82,15 @@
  *
  * <p>The API of this class is subject to change without notice.
  */
-public class SslSocketFactory {
-  private static final Logger logger = Logger.getLogger(SslSocketFactory.class.getName());
+public class CoreSocketFactory {
```
netdpb: Can this class be final?

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
```diff
@@ -151,17 +151,17 @@ public void setup() throws IOException, GeneralSecurityException {
 
   @Test
   public void create_throwsErrorForInvalidInstanceName() throws IOException {
-    SslSocketFactory sslSocketFactory =
-        new SslSocketFactory(new Clock(), clientKeyPair, credential, adminApi, 3307);
+    CoreSocketFactory sslSocketFactory =
```
netdpb: Rename the variable too?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
kurtisvg: Yes - the property is used to the Postgres SF to deal with a backwards compatibility issue  cause by a change in the PGdriver API (it carries the arg under a different key forward to the new key).

The two file formats are passed into the connect method (since the Postgres format requires a suffix). I could create an overloaded method for each, but at some point we may want users to be able to specify a different address, so I thought I would just expose the defaults for now.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
kurtisvg: Done.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
kurtisvg: Removed.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
kurtisvg: Done.

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
kurtisvg: Done.

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -16,64 +16,30 @@
...
+    return socket;
   }
 
   // Cloud SQL sockets always use TLS and the socket returned by connect above is already TLS-ready.
-  // It is fine
-  // to implement these as no-ops.
+  // It is fine to implement these as no-ops.
   @Override
   public Socket beforeHandshake() {
     return socket;
```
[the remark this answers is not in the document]
kurtisvg: My understanding is that it is connect -> beforeHandshake -> afterHandshake.

It's not 100% clear based on the [documentation](https://github.com/mysql/mysql-connector-j/blob/66459e9d39c8fd09767992bc592acd2053279be6/src/legacy/java/com/mysql/jdbc/SocketFactory.java#L69). However our current implementation actually implements the SSL stuff itself in the CoreSocketFactory.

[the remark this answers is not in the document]
netdpb: So they expect the factory to hold onto its single connection. So weird.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
netdpb: OK. We can deal with this in followup PRs, but I'm worried about String parameters that are really more constrained.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -135,13 +135,10 @@
   }
 
   /**
-   * Returns the CoreSocketFactory singleton, which can be used to createSslSocket SslSockets to
-   * Cloud SQL.
-   *
-   * @return the CoreSocketFactory singleton.
+   * Returns the {@link CoreSocketFactory} singleton.
```
netdpb: Make sure to run google-java-format before merging.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -226,7 +223,7 @@ private boolean runningOnGaeStandard() {
    * @throws IOException if error occurs during socket creation.
    */
   // TODO(berezv): separate creating socket and performing connection to make it easier to test
-  public Socket createSslSocket(String instanceName, List<String> ipTypes) throws IOException {
+  protected Socket createSslSocket(String instanceName, List<String> ipTypes) throws IOException {
```
netdpb: Why `protected`? It's a final class.

PR #128, opened 2019-03-07
Replace outdated examples with reference to new samples. 

### LINKED ISSUE #111: Update Example to Follow Best Practice
Hi, I would like to suggest the following sample code to use prepared statement instead. I may try to fork it and do a PR soon.

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/master/examples/compute-engine/src/main/java/com/google/cloud/sql/mysql/example/ListTables.java

kurtisvg:
The samples linked from this repo have been replaced in the doc. This removes the samples and updated a reference to the README to point to a sample application that is more up to date. 

Fixes #111.


README.md
```diff
@@ -10,8 +10,8 @@ whitelisting or SSL certificates manually.
 
 ### Examples
 
-For an examples of a simple application using Cloud SQL Socket Factory, please take a look at the 
-projects found 
+For examples of the SocketFactory being used in the context of an application, check out the sample
```
netdpb:
```suggestion
For examples of `SocketFactory` being used in the context of an application, check out the sample 
```

kurtisvg: I went more generic - I didn't want to accidentally encourage use of the `SocketFactory` class itself.

PR #129, opened 2019-03-29
Release v1.0.13

kurtisvg:
## Release Notes
* Updated dependencies to new versions. 
* Minor cosmetic restructuring for code health.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ are unsure, it is recommended to use the latest version of `mysql-connector-java
 
 | JDBC Driver Version        | Cloud SQL Socket Factory Version         |
 | -------------------------- | ---------------------------------------- |
-| mysql-connector-java:8.x   | mysql-socket-factory-connector-j-8:1.0.12 |
-| mysql-connector-java:6.x   | mysql-socket-factory-connector-j-6:1.0.12 |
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -35,9 +35,9 @@ are unsure, it is recommended to use the latest version of `mysql-connector-java
 
 | JDBC Driver Version        | Cloud SQL Socket Factory Version         |
 | -------------------------- | ---------------------------------------- |
```
netdpb: Is it useful to adjust the length of all rows to be the same?

connector-j-5/pom.xml
```diff
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>com.google.cloud.sql</groupId>
       <artifactId>jdbc-socket-factory-core</artifactId>
-      <version>1.0.13-SNAPSHOT</version>
+      <version>1.0.13</version>
```
netdpb: Can this use `${project.version}`? (You can consider this for the future instead of now.)

Same for everywhere except the references to the parent.

README.md
kurtisvg: Only in that it looks nice in raw MD format

PR #133, opened 2019-04-22
Deprecate USER_TOKEN_PROPERTY_NAME and add method alternatives.

### LINKED ISSUE #131: Relax visibility of CoreSocketFactory.USER_TOKEN_PROPERTY_NAME
`CoreSocketFactory.USER_TOKEN_PROPERTY_NAME` became private in #126. Would it be possible to expose it as a public constant somewhere in the library? We use it in [spring-cloud-gcp](https://github.com/spring-cloud/spring-cloud-gcp/blob/72d746486fedd19b16ead3cdbbcfb48c3e90c3a9/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java#L210) to provide a custom application identity.

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/74aba381161097492eaab917fc36c19a8c59df84/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L617

kurtisvg:
Fixes #131.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -88,9 +88,9 @@ public final class CoreSocketFactory {
   public static final String POSTGRES_SOCKET_FILE_FORMAT = "/cloudsql/%s/.s.PGSQL.5432";
 
   /**
-   * Property used to set the ApplicationName for the underlying SQLAdmin client.
+   * Property used to set the application name for the underlying SQLAdmin client.
    *
-   * @deprecated Replaced by {@link #getApplicationName} and {@link #setApplicationName}.
+   * @deprecated Use {@link #setApplicationName(String)} to set the application name programmatically.
    */
   @Deprecated public static final String USER_TOKEN_PROPERTY_NAME = "_CLOUD_SQL_USER_TOKEN";
 
@@ -731,7 +731,7 @@ public final class CoreSocketFactory {
   /**
    * Sets the User-Agent header for requests made using the underlying SQLAdmin API client.
    *
-   * @throws IllegalStateException if the SQLAdmin client has already been initialized.
+   * @throws IllegalStateException if the SQLAdmin client has already been initialized
    */
   public static void setApplicationName(String applicationName) {
     if (coreSocketFactory != null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -87,10 +87,16 @@
   public static final String MYSQL_SOCKET_FILE_FORMAT = "/cloudsql/%s";
   public static final String POSTGRES_SOCKET_FILE_FORMAT = "/cloudsql/%s/.s.PGSQL.5432";
 
+  /**
+   * Property used to set the ApplicationName for the underlying SQLAdmin client.
```
netdpb:
```suggestion
   * Property used to set the application name for the underlying SQLAdmin client.
```

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -87,10 +87,16 @@
...
+   *
+   * @deprecated Replaced by {@link #getApplicationName} and {@link #setApplicationName}.
```
netdpb:
```suggestion
   * @deprecated Use {@link #setApplicationName(String)} to set the application name programmatically.
```

Most users want to set it. And it's good to use the Javadoc name with parameter types when you can.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -723,6 +728,27 @@ public SSLSocketFactory getSslSocketFactory() {
     }
   }
 
+  /**
+   * Sets the User-Agent header for requests made using the underlying SQLAdmin API client.
+   *
+   * @throws IllegalStateException if the SQLAdmin client has already been initialized.
```
netdpb:
```suggestion
   * @throws IllegalStateException if the SQLAdmin client has already been initialized
```

pom.xml
```diff
@@ -54,6 +54,10 @@
     </developer>
   </developers>
 
+  <properties>
```
netdpb: Is this supposed to be part of this PR?

kurtisvg: I added it to address a warning message that maven logs while compiling - I can move it out into a separate PR if you prefer.

netdpb: No it's fine.

PR #137, opened 2019-05-16
[WIP] Implement Certificate Refresh Asynchronously 

kurtisvg:
Implements asynchronous interaction with the Cloud SQL Admin API.

Future work: 
* Simplify Exception handling
* Evaluate tests
* Enum for IP types


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+  /**
+   * Uses the Cloud SQL Admin API to create an ephemeral SSL certificate that is authenticated to
+   * connect the Cloud SQL instance for 60 minutes.
+   */
+  private Certificate fetchEphemeralCertificate() {
+    // Format the public key into a PEM encoded Certificate.
+    StringBuilder publicKeyPemBuilder = new StringBuilder();
+    publicKeyPemBuilder.append("-----BEGIN RSA PUBLIC KEY-----\n");
+    publicKeyPemBuilder.append(
+        Base64.getEncoder()
```
netdpb: Check out [`BaseEncoding`](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/BaseEncoding.java):

```java
BaseEncoding.base64().withSeparator("\n", 64).encode(...)
```

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    return new Metadata(ipAddrs, instanceCaCertificate);
+  }
+
+  /**
+   * Uses the Cloud SQL Admin API to create an ephemeral SSL certificate that is authenticated to
+   * connect the Cloud SQL instance for 60 minutes.
+   */
+  private Certificate fetchEphemeralCertificate() {
+    // Format the public key into a PEM encoded Certificate.
+    StringBuilder publicKeyPemBuilder = new StringBuilder();
```
netdpb: What about a private method that returns `publicKeyPemBuilder.toString()`?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    } catch (IOException ex) {
+      throw addExceptionContext(
+          ex,
+          String.format(
+              "[%s] Failed to create ephemeral certificate for the Cloud SQL instance.",
+              connectionName));
+    }
+
+    // Parse the certificate from the response.
+    Certificate ephemeralCertificate;
```
netdpb: This looks almost the same as the code in `fetchMetadata()`. Can you extract a method that takes a String representing the bytes of a certificate (`response.getCert()`'s return value)?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    try {
+      byte[] certBytes = response.getCert().getBytes(StandardCharsets.UTF_8);
```
netdpb: Are you sure that the bytes of the certificate are valid UTF8?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+
+  // Schedules task to be executed once the provided futures are complete.
+  private <T> ListenableFuture<T> whenComplete(Callable<T> task, ListenableFuture<?>... futures) {
+    SettableFuture<T> taskFuture = SettableFuture.create();
+
+    // Create a countDown for all Futures to complete.
+    AtomicInteger countDown = new AtomicInteger(futures.length);
+
+    // Trigger the task when all futures are complete.
+    Runnable runOutput =
```
netdpb: Maybe name this `runWhenInputsAreComplete`?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+        };
+    for (ListenableFuture<?> future : futures) {
+      future.addListener(runOutput, executor);
+    }
+
+    return taskFuture;
+  }
+
+  // Returns the inner future from a nested future, or else the exception it encountered.
+  private <T> ListenableFuture<T> extractNestedFuture(
```
netdpb: This can be static to indicate that it's a utility that isn't tied to this instance.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+        ex instanceof GoogleJsonResponseException ? (GoogleJsonResponseException) ex : null;
+    if (gjrEx == null
+        || gjrEx.getDetails() == null
+        || gjrEx.getDetails().getErrors() == null
+        || gjrEx.getDetails().getErrors().isEmpty()) {
+      return new RuntimeException(fallbackDesc, ex);
+    }
+    // Check for commonly occurring user errors and add additional context
+    String reason = gjrEx.getDetails().getErrors().get(0).getReason();
+    if ("accessNotConfigured".equals(reason)) {
```
netdpb: Up to you, but you can use `switch` with string literal cases.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+      // This error occurs when the project doesn't have the "Cloud SQL Admin API" enabled
+      String apiLink =
+          "https://console.cloud.google.com/apis/api/sqladmin/overview?project=" + projectId;
+      return new RuntimeException(
```
netdpb: Shouldn't all these use `ex` as the cause?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+                  + "access it. Please verify the instance connection name and check the IAM "
+                  + "permissions for project \"%s\" ",
+              connectionName, projectId));
+    }
+    // Fallback to the generic description
+    return new RuntimeException(fallbackDesc, ex);
+  }
+
+  /** Represents the results of {@link this.performRefresh}. */
+  private class InstanceData {
```
netdpb:
```suggestion
  private static class InstanceData {
```

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    private Metadata metadata;
```
netdpb: These should both be final.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    }
+
+    Map<String, String> getIpAddrs() {
+      return metadata.getIpAddrs();
+    }
+  }
+
+  /** Represents the results of {@link this.fetchMetadata}. */
+  private class Metadata {
+    private Map<String, String> ipAddrs;
```
netdpb: These should be final.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -230,25 +206,29 @@ private boolean runningOnGaeStandard() {
...
+      String instanceIp = instance.getPreferredIp(ipTypes);
+
+      socket.connect(new InetSocketAddress(instanceIp, serverProxyPort));
+      socket.startHandshake();
+    } catch (Exception ex) {
+      instance.forceRefresh();
+      throw new RuntimeException("Unable to connect to Cloud SQL instance.", ex);
     }
+
+    return socket;
```
netdpb: Return inside the `try` block, and delcare `socket` when you initialize it.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -0,0 +1,72 @@
...
+  @Before
+  public void setup() throws IOException {
+    MockitoAnnotations.initMocks(this);
+
+    when(sqlAdminClient.instances()).thenReturn(sqlAdminClientInstances);
+    when(sqlAdminClientInstances.get(anyString(), anyString()))
+        .thenReturn(sqlAdminClientInstancesGet);
+  }
+
+  @Ignore
```
netdpb: ???

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -0,0 +1,72 @@
...
+      try {
+        // CloudSqlInstance instance = new CloudSqlInstance(test, sqlAdminClient);
+        fail();
+      } catch (IllegalArgumentException ex) {
+        assertThat(ex.getMessage()).contains("Cloud SQL connection name is invalid");
+      }
+    }
+  }
+
+  @Ignore
```
netdpb: ???

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    this.projectId = connFields[0];
+    this.regionId = connFields[1];
+    this.instanceId = connFields[2];
+
+    this.apiClient = apiClient;
+    this.executor = executor;
+    this.keyPair = keyPair;
+
+    // Kick off initial async jobs
+    synchronized (instanceDataGuard) {
```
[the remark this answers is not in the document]
kurtisvg: I almost did that, but I wanted to avoid triggering the rate limiting for the first refresh.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    return (SSLSocket) this.getInstanceData().getSslContext().getSocketFactory().createSocket();
+  }
+
+  /**
+   * Gets the first IP Address for the instance that matches preference.
+   *
+   * @param preferredTypes Preferred types of IP to return. Currently supports "PUBLIC" or
+   *     "PRIVATE".
+   * @return A String representing the IP address of the intance.
+   * @throws IllegalArgumentException If the instance has no IP addresses matching the provided
```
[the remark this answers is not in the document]
kurtisvg: This indicates that the user (of the library) has erred - they provided no preferred types that the instance haves.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+      instanceMetadata = apiClient.instances().get(projectId, instanceId).execute();
+    } catch (IOException ex) {
+      throw addExceptionContext(
+          ex,
+          String.format("[%s] Failed to update metadata for Cloud SQL instance.", connectionName));
+    }
+
+    // Validate the instance will support the authenticated connection.
+    if (!instanceMetadata.getRegion().equals(regionId)) {
+      throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
kurtisvg: So in this case, this implies that the argument to the _socketFactory_ is bad - the user has provided an incorrect connection name.

I really have no idea - this is left over behavior from the previous version. It might be possible to have multiple instances with the same instance name but different regions, but I'm not positive.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+          String.format(
+              "[%s] The region specified for the Cloud SQL instance is"
+                  + " incorrect. Please verify the instance connection name.",
+              connectionName));
+    }
+    if (!instanceMetadata.getBackendType().equals("SECOND_GEN")) {
+      throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
kurtisvg: I choose `IllegalArgumentException` for the same reason - it implies the user provided an incorrect argument to the SocketFactory via the instance connection name.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
kurtisvg: I'm not sure - this is behavior that was converted over from [here](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/eaa61edeea8755f07eef15d5177ed8674b375bea/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L597). 

AFAIK the API docs don't specify: https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/sslCerts#resource

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
kurtisvg: I didn't use it here because I'm not certain reason won't be null safe.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+    }
+    return sslContext;
+  }
+
+  /** Fetches the latest version of the instance's metadata using the Cloud SQL Admin API. */
+  private Metadata fetchMetadata() {
+    DatabaseInstance instanceMetadata;
+    try {
+      instanceMetadata = apiClient.instances().get(projectId, instanceId).execute();
+    } catch (IOException ex) {
```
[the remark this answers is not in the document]
kurtisvg: I'm not sure I understand - the catch should be at the bottom of this entire function?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,496 @@
...
+  /**
+   * Gets the first IP Address for the instance that matches preference.
+   *
+   * @param preferredTypes Preferred types of IP to return. Currently supports "PUBLIC" or
+   *     "PRIVATE".
+   * @return A String representing the IP address of the intance.
+   * @throws IllegalArgumentException If the instance has no IP addresses matching the provided
+   *     preferences.
+   */
+  String getPreferredIp(List<String> preferredTypes) {
```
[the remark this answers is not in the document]
kurtisvg: Possibly? [Here](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/78#discussion_r217852687) is the discussion on it from when it was originally added.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -140,92 +144,63 @@ String getPreferredIp(List<String> preferredTypes) {
...
+      return false;
     }
     synchronized (instanceDataGuard) {
-      // If a scheduled Refresh already exists, cancel if if not immediately imminent
-      if (nextInstanceData == null || nextInstanceData.cancel(false)) {
-        // If canceled successfully, schedule a replacement immediately
-        nextInstanceData = null;
-        scheduleRefresh(0, TimeUnit.MINUTES);
+      // If a scheduled refresh exists and can't be canceled, fail the operation
+      if (nextInstanceData != null && !nextInstanceData.cancel(false)) {
```
netdpb: There's still the case where someone calls `forceRefresh()` while the initial future is pending.

1. The constructor calls `forceRefresh()`. `nextInstanceData` is `null` until that refresh completes.
3. Before that refresh completes, someone calls `forceRefresh()`.
4. Since `nextInstanceData` is still `null`, the `if` condition is false, so `currentInstanceData` is replaced by a new incomplete future.
5. The first operation completes, and replaces `currentInstanceData` with the first completed future.
6. The second operation completes, and replaces `currentInstanceData` with the second completed future.

That will be okay for the users; it shouldn't mean any extra delays. But it does mean more than one pending operation at once.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -97,6 +101,7 @@
    */
   private InstanceData getInstanceData() {
     try {
+      // TODO(kvg): Let exceptions up to here before adding context
       return Uninterruptibles.getUninterruptibly(currentInstanceData);
```
netdpb: This read must be synchronized with `instanceDataGuard` as well as the writes.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -53,8 +54,8 @@
   private final KeyPair keyPair;
 
   private final Object instanceDataGuard = new Object();
-  private ListenableFuture<InstanceData> currentInstanceData;
-  private ListenableFuture<ListenableFuture<InstanceData>> nextInstanceData;
+  private volatile ListenableFuture<InstanceData> currentInstanceData;
```
netdpb: I'm no longer convinced these need to be volatile since they're covered by synchronized blocks.

netdpb: But consider annotating both with [`@GuardedBy("instanceDataGuard")`](https://github.com/google/error-prone/blob/master/annotations/src/main/java/com/google/errorprone/annotations/concurrent/GuardedBy.java). That should help you avoid unsynchronized access.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
netdpb: An alternative is to initialize `nextInstanceData` with a completed future, maybe `Futures.immediateFuture(null)`. Then it'll never be `null`, and in that state won't be cancellable.

kurtisvg: Would `Futures.immediateCancelledFuture()` make more sense than `Futures.immediateFuture(null)`?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -340,27 +354,60 @@ private String generatePublicKeyCert() {
...
     for (ListenableFuture<?> future : futures) {
-      future.addListener(runWhenInputsAreComplete, executor);
+      Futures.addCallback(future, runWhenInputAreComplete, executor);
     }
 
     return taskFuture;
   }
 
+  /** Returns a future that blocks until the result of a nested future is complete. */
+  private <T> ListenableFuture<T> blockOnNestedFuture(
```
netdpb: Can this be static and take Executor to indicate that it's a pure function?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -100,13 +113,15 @@
...
    */
   private InstanceData getInstanceData() {
-    try {
-      // TODO(kvg): Let exceptions up to here before adding context
-      return Uninterruptibles.getUninterruptibly(currentInstanceData);
-    } catch (ExecutionException ex) {
-      Throwable cause = ex.getCause();
-      Throwables.throwIfUnchecked(cause);
-      throw new RuntimeException(cause);
+    synchronized (instanceDataGuard) {
```
netdpb: I think if you synchronize inside the try, you'll release the lock earlier on an exception. Probably doesn't matter in practice, but generally synchronization blocks should be as small as possible.

Now that I think about it, you actually don't want to synchronize while waiting at all. Try this:

```java
ListenableFuture<InstanceData> instanceData;
  instanceData = currentInstanceData;
}
try {
 return getUninterruptibly(instanceData);
} catch (...) { ... }
```

Now you've released the lock before you block.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -91,7 +103,8 @@
 
     // Kick off initial async jobs
     synchronized (instanceDataGuard) {
-      this.currentInstanceData = this.performRefresh();
+      this.nextInstanceData = Futures.immediateFuture(performRefresh());
```
netdpb: OK. What about this?

```java
this.currentInstanceData = performRefresh();
this.nextInstanceData = immediateFuture(currentInstanceData);
```

Now you don't need to call `blockOnNestedFuture` here.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -154,13 +169,12 @@ boolean forceRefresh() {
       return false;
     }
     synchronized (instanceDataGuard) {
-      // If a scheduled refresh exists and can't be canceled, fail the operation
-      if (nextInstanceData != null && !nextInstanceData.cancel(false)) {
-        return false;
+      // If a scheduled refresh doesn't exist or hasn't started, perform one immediately
+      if (nextInstanceData == null || nextInstanceData.cancel(false)) {
+        nextInstanceData = Futures.immediateFuture(performRefresh());
```
netdpb: Oh, I like this!

Could you avoid the `blockOnNestedFuture` call in this case, too?

```java
if (nextInstanceData == null || nextInstanceData.cancel(false)) {
  currentInstanceData = performRefresh();
  nextInstanceData = immediateFuture(currentInstanceData);
} else {
  currentInstanceData = blockOnNestedFuture(nextInstanceData);
}
```

Now you have only one call to `blockOnNestedFuture`, because the other cases are simpler.

And since those two lines in the if block are the same as what you do in the constructor, you could have a simple `replaceCurrentInstanceData()` method that's called from both places, if you want.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -340,27 +354,60 @@ private String generatePublicKeyCert() {
   }
 
   // Schedules task to be executed once the provided futures are complete.
-  private <T> ListenableFuture<T> whenAllComplete(
-      Callable<T> task, ListenableFuture<?>... futures) {
+  private <T> ListenableFuture<T> whenAllSucceed(Callable<T> task, ListenableFuture<?>... futures) {
```
netdpb: Can you make this static and take the Executor as a parameter, since it's really meant to be a pure function?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -340,27 +354,60 @@ private String generatePublicKeyCert() {
...
 
     // Trigger the task when all futures are complete.
-    Runnable runWhenInputsAreComplete =
-        () -> {
-          if (countDown.decrementAndGet() == 0) {
-            taskFuture.setFuture(executor.submit(task));
+    FutureCallback<Object> runWhenInputAreComplete =
+        new FutureCallback<Object>() {
+          @Override
+          public void onSuccess(@NullableDecl Object o) {
```
netdpb: Are you using the checker framework? Or did IntelliJ put the `@NullableDecl` there?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -340,27 +354,60 @@ private String generatePublicKeyCert() {
...
+            if (countDown.decrementAndGet() == 0) {
+              taskFuture.setFuture(executor.submit(task));
+            }
+          }
+
+          @Override
+          public void onFailure(Throwable throwable) {
+            if (!taskFuture.setException(throwable)) {
+              String msg =
+                  "Got more than one input Future Failure. Logging failures after the first";
```
netdpb: Nit: I'd leave out "Future" here, and lowercase "failure".

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
kurtisvg: IntelliJ did it. Should I remove?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
kurtisvg: IIRC, I copied verbatim from the guava line

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
kurtisvg: In this case, do we even need to synchronize? I thought references assignments were already atomic in java, so doesn't calling the function already use the value of the reference?

netdpb: They are if you make the field `volatile`.

If you're using synchronized blocks, you really should synchronize around reads and writes. And since you need to synchronize across two fields, `volatile` isn't enough. I was hoping we could get down to one field so that you could just make it `volatile`, but we haven't.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
netdpb: If you're using the checker framework, then we can apply those annotations across the implementation. If not, they're noise IMO.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -0,0 +1,509 @@
...
+   *
+   * @return {@code true} if successfully scheduled, or {@code false} otherwise.
+   */
+  boolean forceRefresh() {
+    if (!forcedRenewRateLimiter.tryAcquire()) {
+      return false;
+    }
+    synchronized (instanceDataGuard) {
+      // If a scheduled refresh doesn't exist or hasn't started, perform one immediately
+      if (nextInstanceData == null || nextInstanceData.cancel(false)) {
```
netdpb: I don't think `nextInstanceData` is ever `null` now.

```suggestion
      if (nextInstanceData.cancel(false)) {
```

netdpb:
OK, except for the capital Failure. As I said, this is a tiny nit.

kurtisvg:
> OK, except for the capital Failure. As I said, this is a tiny nit.

Should be just `"Got more than one input failure. Logging failures after the first"` currently..

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -230,24 +206,28 @@ private boolean runningOnGaeStandard() {
...
+      socket.setKeepAlive(true);
+      socket.setTcpNoDelay(true);
+
+      String instanceIp = instance.getPreferredIp(ipTypes);
+
+      socket.connect(new InetSocketAddress(instanceIp, serverProxyPort));
+      socket.startHandshake();
+
+      return socket;
+    } catch (Exception ex) {
```
netdpb: TODO for later: Catch `IOException` only, and move the `forceRefresh()` call into `getInstanceData()`.

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
```diff
@@ -179,12 +177,13 @@ public void create_throwsErrorForInvalidInstanceRegion() throws IOException {
...
 
     CoreSocketFactory coreSocketFactory =
-        new CoreSocketFactory(new Clock(), clientKeyPair, credential, adminApi, 3307);
+        new CoreSocketFactory(clientKeyPair, credential, adminApi, 3307);
     try {
       coreSocketFactory.createSslSocket("foo:bar:baz", Arrays.asList("PRIMARY"));
       fail();
     } catch (IllegalArgumentException e) {
-      assertThat(e.getMessage()).startsWith("Incorrect region value");
+      assertThat(e.getMessage())
```
netdpb:
```suggestion
      assertThat(e).hasMessageThat()
```

Do this throughout, and you'll get better failure messages.

PR #144, opened 2019-06-14
Set executor shutdown time to 0 to prevent application hanging.

kurtisvg:


netdpb:
Exiting executors should quit immediately unless they're currently running something. But scheduled executors won't (by default) if they have unstarted tasks. To make the executor cancel unstarted tasks on exit, I think you should call [`executor.setExecuteExistingDelayedTasksAfterShutdownPolicy(false)`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html#setExecuteExistingDelayedTasksAfterShutdownPolicy-boolean-) instead.

PR #146, opened 2019-06-21
Remove unnecessary instance initialization. 

kurtisvg:
Remove unnecessary instance initialization from postgres module.

### DIFF
--- a/postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java
+++ b/postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java
@@ -68,8 +68,7 @@ public class SocketFactory extends javax.net.SocketFactory {
 
   @Override
   public Socket createSocket() throws IOException {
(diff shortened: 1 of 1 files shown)


kurtisvg:
> Interesting. Did Error-Prone catch this for you?

Unfortunately not - a user for GCF noticed RSA keys were still being generated for Postgres.

hravnaas:
Hi, @kurtisvg,  any idea when this will be released, so version 1.0.15 I assume?

kurtisvg:
@hravnaas Yes, I'm currently to release around every 2 months, so the next release should be in a couple of weeks (first week of august) . 

You should be able to use a snapshot version to deploy if you clone this repo and build with `mvn install` before deploying:
```
<dependency>
    <artifactId>mysql-socket-factory-connector-j-8</artifactId>
    <version>1.0.15-SNAPSHOT</version>
</dependency>
```

PR #150, opened 2019-07-25
Support legacy domain-scoped projects

aniket486:
For example - google.com:project:region:instance-name

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -86,18 +86,23 @@ class CloudSqlInstance {
       ListeningScheduledExecutorService executor,
       ListenableFuture<KeyPair> keyPair) {
     String[] connFields = connectionName.split(":");
-    if (connFields.length != 3) {
+    if (connFields.length == 3) {
+      this.projectId = connFields[0];
+      this.regionId = connFields[1];
+      this.instanceId = connFields[2];
+    } else if (connFields.length == 4) {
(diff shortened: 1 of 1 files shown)


kurtisvg:
I was unaware that project names could even contain colons.

This doesn't really fix the problem, only one use case- if the user has multiple colons in the project name they will have the same problem.

kurtisvg:
Digging into this a little more, it looks like project ids with a colon are "domain scoped", which is a legacy feature. It looks these projects only allow for one colon, so this fix could be suitable. 

I need to confirm there isn't any existing precedence for this first.

aniket486:
@kurtisvg - any update on this?
You are right, there are a few legacy projects that have google.com: as a prefix. We need this change in order to support such projects.

kurtisvg:
Not yet - it looks like support for this is varied by product. [GCR substitutes with a slash](https://cloud.google.com/container-registry/docs/overview#domain-scoped_projects), Cloud Run just declares itself incompatible with domain scoped projects.

I'm waiting to hear back from the Cloud SQL team -- I want to make sure that whatever path we take is consistent with the proxy and Cloud SQL in other areas.

kurtisvg:
It looks like the domain might be scoped only to the project name, and not the project_id. You should be able to already use the library with the string provided on the Cloud SQL instance's overview page. Can you confirm that this works for you?

kurtisvg:
Ping on this - can I close this PR out or is it still needed?

kurtisvg:
Just found another occurrence of a user that ran into this issue, so I'm re-opening for now. Using # 165 to track.

PR #151, opened 2019-07-26
Using google-auth-library-oauth2-http to get default credentials

yhrn:
This should enable use of [Workload Identities](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) on GKE, which requires [this fix](https://github.com/googleapis/google-auth-library-java/pull/283) included in google-auth-library-oauth2-http 0.16.2

The solution isn't terribly beautiful but it seemed like a simple way to do it while maintaining backwards compatibility.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/CredentialFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/CredentialFactory.java
@@ -17,11 +17,12 @@
 package com.google.cloud.sql;
 
 import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.http.HttpRequestInitializer;
 
 /** Factory for creating {@link Credential}s for interaction with Cloud SQL Admin API. */
 public interface CredentialFactory {
   /** Name of system property that can specify an alternative credential factory. */
   String CREDENTIAL_FACTORY_PROPERTY = "cloudSql.socketFactory.credentialFactory";
 
-  Credential create();
+  HttpRequestInitializer create();
 }
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -16,8 +16,6 @@
 
 package com.google.cloud.sql.core;
 
-import com.google.api.client.auth.oauth2.Credential;
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.HttpTransport;
@@ -26,6 +24,8 @@ import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.services.sqladmin.SQLAdmin;
 import com.google.api.services.sqladmin.SQLAdmin.Builder;
 import com.google.api.services.sqladmin.SQLAdminScopes;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.sql.CredentialFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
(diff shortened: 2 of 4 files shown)


kurtisvg:
Sorry, this is on my radar but I haven't had a time to dive very deep on it right now. 

My two current concerns are: 
1. if this will have any effect on the spring-integration-gcp, which relies on this project. 
2. if we actually need to add a pre 1.0 dependency with `google-auth-library-oauth2-http` 

@meltsufin - Any concerns about this from the `spring-integration-gcp` side?

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
```diff
@@ -17,11 +17,12 @@
...
 import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.http.HttpRequestInitializer;
 
 /** Factory for creating {@link Credential}s for interaction with Cloud SQL Admin API. */
 public interface CredentialFactory {
   /** Name of system property that can specify an alternative credential factory. */
   String CREDENTIAL_FACTORY_PROPERTY = "cloudSql.socketFactory.credentialFactory";
 
-  Credential create();
+  HttpRequestInitializer create();
```
meltsufin: This is a breaking changing to a public interface.
I would expect a major version bump for this to permissible.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -321,19 +296,19 @@ private static SQLAdmin createAdminApiClient(HttpRequestInitializer requestIniti
...
       }
-      if (credential.createScopedRequired()) {
-        credential =
-            credential.createScoped(Collections.singletonList(SQLAdminScopes.SQLSERVICE_ADMIN));
+      if (credentials.createScopedRequired()) {
+        credentials =
+            credentials.createScoped(Collections.singletonList(SQLAdminScopes.SQLSERVICE_ADMIN));
       }
-      return credential;
+      return new HttpCredentialsAdapter(credentials);
```
meltsufin: So why not make this method return `com.google.auth.Credentials` instead and do the wrapping in the caller?
You would change the method signature to `Credentials create()` as well.

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
yhrn: It might break binary compatibility, but it's source compatible

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
yhrn: That would be a source code breaking change, `com.google.api.client.auth.oauth2.Credential` implements `com.google.api.client.http.HttpRequestInitializer` so this way the PR is source code compatible - existing implementations of `com.google.cloud.sql.CredentialFactory` will not need any source code changes, only be recompiled against the new library version. `com.google.auth.Credentials` on the other hand is a completely separate API and would require all existing implementations to change.

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
meltsufin: I didn't realize `Credential` implements `HttpRequestInitializer`.  This is actually OK then.

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
```diff
@@ -17,11 +17,12 @@
 package com.google.cloud.sql;
 
 import com.google.api.client.auth.oauth2.Credential;
```
meltsufin: Unused import?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
meltsufin: You right. Thanks for the clarification.

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
yhrn: It's actually used in the comment but maybe the comment should be updated? I left it as is because I figured that in most cases it would be a `Credential` that is returned.

meltsufin: Ah, yeah that's fine then.

PR #153, opened 2019-07-30
Update TravisCI to openjdk.

kurtisvg:
Travis CI has switch the default build environment, which means that `oraclejdk8` is no longer available. Switching to openjdk and adding a few other supported versions.


kurtisvg:
Update - openjdk11 doesn't work correctly with travis, despite working locally. I've reverted for now and will add multiple versions when we migrate to Kokoro.

PR #163, opened 2019-11-06
Clarify host parameter is optional in the README.

kurtisvg:


README.md
```diff
@@ -96,6 +96,8 @@ The full JDBC url should look like this:
 jdbc:mysql://google/<DATABASE_NAME>?cloudSqlInstance=<INSTANCE_CONNECTION_NAME>&socketFactory=com.google.cloud.sql.mysql.SocketFactory&useSSL=false&user=<MYSQL_USER_NAME>&password=<MYSQL_USER_PASSWORD>
```
saturnism: can we remove the `google` host here too?

README.md
```diff
@@ -96,6 +96,8 @@ The full JDBC url should look like this:
...
 ```
 
+Note: The host portion of the JDBC url is currently unused, and has no effect on the connection process. The SocketFactory will get your instances IP address base on the provided `cloudSqlInstance` arg.
```
saturnism: url -> URL

saturnism: also, is this comment specific to MySQL? What about PostgresSQL?

README.md
```diff
@@ -96,6 +96,8 @@ The full JDBC url should look like this:
 jdbc:mysql://google/<DATABASE_NAME>?cloudSqlInstance=<INSTANCE_CONNECTION_NAME>&socketFactory=com.google.cloud.sql.mysql.SocketFactory&useSSL=false&user=<MYSQL_USER_NAME>&password=<MYSQL_USER_PASSWORD>
 ```
 
+Note: The host portion of the JDBC url is currently unused, and has no effect on the connection process. The SocketFactory will get your instances IP address base on the provided `cloudSqlInstance` arg. 
+
 ##### Postgres
 
 Base JDBC url: `jdbc:postgresql://google/<DATABASE_NAME>`
```
saturnism: this should also be consistent

README.md
kurtisvg: Done.

README.md
kurtisvg: Done.

README.md
kurtisvg: Done.

README.md
saturnism: while we are at it, can we just upper case the url? :D

kurtisvg: @saturnism - I don't see any that are still lowercase - do you see one in the current version? I was having some issues with Chrome/Github's search function stepping on each other, but I in the page preview they are all uppercase'd.

saturnism: i see them now. thnx!

PR #168, opened 2020-01-29
Moved packaging of drivers for Cloud Data Fusion to a separate profile

bdmogal:


bdmogal:
@kurtisvg could you PTAL?

pom.xml
```diff
@@ -237,6 +237,34 @@
...
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-assembly-plugin</artifactId>
+            <version>2.4.1</version>
+            <configuration>
+              <!-- get all project dependencies -->
+              <descriptorRefs>
+                <descriptorRef>jar-with-dependencies</descriptorRef>
```
kurtisvg: nit: Should this be `jar-with-driver-and-dependencies`

bdmogal: yeah, I tried that, but didn't work. Apparently this is not free-form text. Here's the error I got:
{code}
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.4.1:single (make-assembly) on project jdbc-socket-factory-parent: Error reading assemblies: Descriptor with ID 'jar-with-driver-and-dependencies' not found -> [Help 1]
{code}

kurtisvg: Ah yup: https://maven.apache.org/plugins/maven-assembly-plugin/descriptor-refs.html#jar-with-dependencies

Ok, thanks

bdmogal:
@kurtisvg thanks for merging this. Can we plan on doing a release to make this available to users?

kurtisvg:
Releasing a new version doesn't really make sense as there aren't any changes since v1.15. I can just add some jar-with-drivers-and-dependencies to the v1.15 release - will that work?

bdmogal:
Yes, as long as it is hosted somewhere, its good enough.

kurtisvg:
Ok, added to the releases page: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/releases/tag/v1.0.15

Please verify they work correctly for Cloud Data Fusion.

PR #210, opened 2020-04-08
Improve handling of connections through unix domain sockets. 

### LINKED ISSUE #172: Support Cloud Run for Anthos
I'm deploying a Spring Boot application on Cloud Run For Anthos (on my own GKE cluster). As such, there is no unix socket created for Cloud SQL connections. But the logic in CoreSocketFactory assumes that this is a Cloud Function and therefore forces unix socket mode which does not work. Specifically, the logic assumes that the ENV vars K_REVISION and K_SERVICE are only set for Cloud Functions. But these env vars are also set by Cloud Run for Anthos. As such, it is impossible to use this library on Cloud Run for Anthos as there is no way to force it not to revert to unix socket mode.

kurtisvg:
- Adds new `unixSocket` property, which which can be used to specify a path to a unix socket that will connect to an instance
- Deprecates `"CLOUD_SQL_FORCE_UNIX_SOCKET"` env var in favor of the above property
- Disables the use of the `/cloudsql` unix socket by default when on GAE or GCF

Fixes #172

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,9 +160,16 @@ this purpose.
  
 ### Connection via Unix Sockets
 
-The library will automatically detect when it is running on GAE Standard, and will connect via the 
- provided unix socket for reduced latency.
+To connect using a Unix domain socket (such as created by the Cloud SQL proxy), 
+you can use the `"unixSocketPath"` property to specify a path to a local file 
+instead of connecting directly over TCP.
 
-To force the library to connect to a unix socket (typically created by the Cloud SQL proxy) when 
-running outside of the GAE-Standard environment, set the environment variable 
-`CLOUD_SQL_FORCE_UNIX_SOCKET` to any value.
+Example using MySQL:
+```
+jdbc:mysql:///<DATABASE_NAME>?unixSocketPath=</PATH/TO/UNIX/SOCKET>&cloudSqlInstance=<INSTANCE_CONNECTION_NAME>&socketFactory=com.google.cloud.sql.mysql.SocketFactory&user=<MYSQL_USER_NAME>&password=<MYSQL_USER_PASSWORD>
+```
+
+Example using PostgreSQL:
+```
+jdbc:postgresql:///<DATABASE_NAME>?unixSocketPath=</PATH/TO/UNIX/SOCKET>&cloudSqlInstance=<INSTANCE_CONNECTION_NAME>&socketFactory=com.google.cloud.sql.postgres.SocketFactory&user=<POSTGRESQL_USER_NAME>&password=<POSTGRESQL_USER_PASSWORD>
+```
--- a/README.md
+++ b/README.md
@@ -160,9 +160,9 @@ this purpose.
  
 ### Connection via Unix Sockets
 
-To connect using a Unix domain socket (such as created by the Cloud SQL proxy), 
-you can use the `"unixSocketPath"` property to specify a path to a local file 
-instead of connecting directly over TCP.
+To connect using a Unix domain socket (such as the one created by the Cloud SQL 
+proxy), you can use the `unixSocketPath` property to specify a path to a local 
+file instead of connecting directly over TCP.
 
 Example using MySQL:
 ```
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -116,7 +116,7 @@ public final class CoreSocketFactory {
                   + " '%s=\"/cloudsql/INSTANCE_CONNECTION_NAME\"' property in your JDBC url"
                   + " instead.",
               UNIX_SOCKET_PROPERTY));
-      if (prop.get(UNIX_SOCKET_PROPERTY) != null) {
+      if (prop.getProperty(UNIX_SOCKET_PROPERTY) != null) {
         String instanceName = prop.getProperty(CLOUD_SQL_INSTANCE_PROPERTY, "");
         prop.setProperty(UNIX_SOCKET_PROPERTY, "/cloudsql/" + instanceName);
       }
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -63,7 +63,7 @@ import jnr.unixsocket.UnixSocketChannel;
  */
 public final class CoreSocketFactory {
   public static final String CLOUD_SQL_INSTANCE_PROPERTY = "cloudSqlInstance";
-  public static final String UNIX_SOCKET_PROPERTY = "unixSocket";
+  public static final String UNIX_SOCKET_PROPERTY = "unixSocketPath";
 
   /**
    * Property used to set the application name for the underlying SQLAdmin client.
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -63,7 +63,7 @@ import jnr.unixsocket.UnixSocketChannel;
  */
 public final class CoreSocketFactory {
   public static final String CLOUD_SQL_INSTANCE_PROPERTY = "cloudSqlInstance";
-  public static final String UNIX_SOCKET_PROPERTY = "unixSocketPath";
+  private static final String UNIX_SOCKET_PROPERTY = "unixSocketPath";
 
   /**
    * Property used to set the application name for the underlying SQLAdmin client.
@@ -170,8 +170,6 @@ public final class CoreSocketFactory {
 
   /**
    * Creates a socket representing a connection to a Cloud SQL instance.
-   *
-   * @see CoreSocketFactory#connect(Properties, String)
    */
   public static Socket connect(Properties props) throws IOException {
     return connect(props, null);
@@ -183,14 +181,13 @@ public final class CoreSocketFactory {
    * <p>Depending on the given properties, it may return either a SSL Socket or a Unix Socket.
    *
    * @param props Properties used to configure the connection.
-   * @param unixPathSuffix Optional suffix to add the the Unix socket path. Unused if null.
+   * @param unixPathSuffix suffix to add the the Unix socket path. Unused if null.
    * @return the newly created Socket.
    * @throws IOException if error occurs during socket creation.
    */
   public static Socket connect(Properties props, String unixPathSuffix) throws IOException {
     // Gather parameters
     final String csqlInstanceName = props.getProperty(CLOUD_SQL_INSTANCE_PROPERTY);
-    final List<String> ipTypes = listIpTypes(props.getProperty("ipTypes", DEFAULT_IP_TYPES));
 
     // Validate parameters
     Preconditions.checkArgument(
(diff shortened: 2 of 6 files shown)


kurtisvg:
Friendly ping on this.

netdpb:
> Friendly ping on this.

Whoops, sorry!

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -171,12 +188,13 @@ public static Socket connect(Properties props, String socketPathFormat) throws I
             + "connection Properties with value in form \"project:region:instance\"");
 
     // GAE Standard + GCF provide a connection path at "/cloudsql/<CONNECTION_NAME>"
-    if (forceUnixSocket || runningOnGaeStandard() || runningOnGoogleCloudFunctions()) {
+    String unixSocket = props.getProperty(UNIX_SOCKET_PROPERTY);
```
netdpb, line 191: It looks like all the callers to `migrateForceSocketVarToProperty(props)` immediately call this method `connect(props)`. So why not do the migration here, once? And then you don't even need to mutate the properties. You could just have a method like `getUnixSocket(props)` that returns the unix socket property, or the /cloudsql/instance string if the environment variable is set, and appends the suffix if needed.

```java
String unixSocket = getUnixSocket(props);
if (unixSocket != null) {
  logger.info(…);
  return UnixSocketChannel.open…
}
```

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -105,6 +104,25 @@
...
+  public static void migrateForceSocketVarToProperty(Properties prop) {
+    if (System.getenv("CLOUD_SQL_FORCE_UNIX_SOCKET") != null) {
+      logger.warning(
+          String.format(
+              "\"CLOUD_SQL_FORCE_UNIX_SOCKET\" env var has been deprecated. Please use"
+                  + " '%s=\"/cloudsql/INSTANCE_CONNECTION_NAME\"' property in your JDBC url"
+                  + " instead.",
+              UNIX_SOCKET_PROPERTY));
+      if (prop.get(UNIX_SOCKET_PROPERTY) != null) {
+        String instanceName = prop.getProperty(CLOUD_SQL_INSTANCE_PROPERTY, "");
```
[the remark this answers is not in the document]
kurtisvg: We already throw an error in the connect method:
https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/6579df05e078a0f76241b939d79dda1ab2bc17ad/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L185-L188

 before the unix socket is dialed. I didn't want to duplicate.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
kurtisvg: So my goal was to try and encapsulate the Postges specific logic inside the Postgres module. It unfortunately has the suffix that has to be added:

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/6579df05e078a0f76241b939d79dda1ab2bc17ad/postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java#L55-L62

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -105,6 +104,25 @@
...
+   */
+  public static void migrateForceSocketVarToProperty(Properties prop) {
+    if (System.getenv("CLOUD_SQL_FORCE_UNIX_SOCKET") != null) {
+      logger.warning(
+          String.format(
+              "\"CLOUD_SQL_FORCE_UNIX_SOCKET\" env var has been deprecated. Please use"
+                  + " '%s=\"/cloudsql/INSTANCE_CONNECTION_NAME\"' property in your JDBC url"
+                  + " instead.",
+              UNIX_SOCKET_PROPERTY));
+      if (prop.get(UNIX_SOCKET_PROPERTY) != null) {
```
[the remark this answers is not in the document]
kurtisvg: Done.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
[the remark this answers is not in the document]
netdpb: OK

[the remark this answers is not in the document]
netdpb: OK.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
netdpb: Oh I see. What about taking that suffix and passing it to an overload of `connect()`? Then the PostGres factory could call `connect(props, POSTGRES_SUFFIX)` and all the others could call `connect(props)` (where `connect(props)` would just call `connect(props, "")`).

That lets you keep the Postgres suffix in that module but do the environment-variable thing only once.

kurtisvg: I consider that vs my current approach and didn't really have a strong preference for one or the other. I'll refactor to that way if you think it's more reasonable.

kurtisvg: Done.

README.md
```diff
@@ -160,9 +160,16 @@ this purpose.
  
 ### Connection via Unix Sockets
 
-The library will automatically detect when it is running on GAE Standard, and will connect via the 
- provided unix socket for reduced latency.
+To connect using a Unix domain socket (such as created by the Cloud SQL proxy),
```
netdpb, line 163:
```suggestion
To connect using a Unix domain socket (such as the one created by the Cloud SQL proxy), 
```

README.md
```diff
@@ -160,9 +160,16 @@ this purpose.
  
 ### Connection via Unix Sockets
 
-The library will automatically detect when it is running on GAE Standard, and will connect via the 
- provided unix socket for reduced latency.
+To connect using a Unix domain socket (such as created by the Cloud SQL proxy), 
+you can use the `"unixSocketPath"` property to specify a path to a local file
```
netdpb, line 164: I think you only need "" or ``, not both.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -167,16 +148,46 @@ static ListeningScheduledExecutorService getDefaultExecutor() {
...
+    if (unixSocketPath != null) {
+      // Get the Unix socket file path from the properties object
+      return unixSocketPath;
+    } else if (System.getenv("CLOUD_SQL_FORCE_UNIX_SOCKET") != null) {
+      // If the deprecated env var is set, warn and use `/cloudsql/INSTANCE_CONNECTION_NAME`
+      // A socket factory is provided at this path for GAE, GCF, and Cloud Run
+      logger.warning(
+          String.format(
+              "\"CLOUD_SQL_FORCE_UNIX_SOCKET\" env var has been deprecated. Please use"
+                  + " '%s=\"/cloudsql/INSTANCE_CONNECTION_NAME\"' property in your JDBC url"
```
netdpb, line 163: I don't understand the reference to a URL here. Aren't you asking them to set a property?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -167,16 +148,46 @@ static ListeningScheduledExecutorService getDefaultExecutor() {
...
+      return "/cloudsql/" + props.getProperty(CLOUD_SQL_INSTANCE_PROPERTY);
+    }
+    return null; // if unset, default to null
+  }
+
   /**
    * Creates a socket representing a connection to a Cloud SQL instance.
    *
-   * <p>Depending on the environment, it may return either a SSL Socket or a Unix Socket.
+   * @see CoreSocketFactory#connect(Properties, String)
```
netdpb, line 174: I don't think your users need this `@see`.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -167,16 +148,46 @@ static ListeningScheduledExecutorService getDefaultExecutor() {
...
+    return connect(props, null);
+  }
+
+  /**
+   * Creates a socket representing a connection to a Cloud SQL instance.
+   *
+   * <p>Depending on the given properties, it may return either a SSL Socket or a Unix Socket.
    *
    * @param props Properties used to configure the connection.
+   * @param unixPathSuffix Optional suffix to add the the Unix socket path. Unused if null.
```
netdpb, line 186:
```suggestion
   * @param unixPathSuffix suffix to add the the Unix socket path. Unused if null.
```

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -167,16 +148,46 @@ static ListeningScheduledExecutorService getDefaultExecutor() {
...
    * @return the newly created Socket.
    * @throws IOException if error occurs during socket creation.
    */
-  public static Socket connect(Properties props) throws IOException {
+  public static Socket connect(Properties props, String unixPathSuffix) throws IOException {
     // Gather parameters
     final String csqlInstanceName = props.getProperty(CLOUD_SQL_INSTANCE_PROPERTY);
     final List<String> ipTypes = listIpTypes(props.getProperty("ipTypes", DEFAULT_IP_TYPES));
```
netdpb, line 193: You could move this down until after the unix socket block, since you don't use it until then.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -63,8 +63,7 @@
  */
 public final class CoreSocketFactory {
   public static final String CLOUD_SQL_INSTANCE_PROPERTY = "cloudSqlInstance";
-  public static final String MYSQL_SOCKET_FILE_FORMAT = "/cloudsql/%s";
-  public static final String POSTGRES_SOCKET_FILE_FORMAT = "/cloudsql/%s/.s.PGSQL.5432";
+  public static final String UNIX_SOCKET_PROPERTY = "unixSocketPath";
```
netdpb, line 66: Does this have to be public? I don't think it's used anywhere except in this file.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
kurtisvg: The JDBC specification uses a url to represent database connections, for example:

`jdbc:postgresql:///<DATABASE_NAME>?unixSocketPath=</PATH/TO/UNIX/SOCKET>&cloudSqlInstance=<INSTANCE_CONNECTION_NAME>`

The driver parses it into a bunch of objects that are passed into the ConnectionFactory, including the Properties object from the query string.

netdpb: Got it.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -213,6 +210,7 @@ public static Socket connect(Properties props, String unixPathSuffix) throws IOE
       return UnixSocketChannel.open(socketAddress).socket();
     }
 
+    final List<String> ipTypes = listIpTypes(props.getProperty("ipTypes", DEFAULT_IP_TYPES));
```
netdpb, line 213: Not important for this PR, but generally you don't need to mark local variables `final`. It's considered more noise than the extra information is worth.

PR #225, opened 2020-06-11
Update User Agent methods and format

### LINKED ISSUE #219: Update UserAgent format for usage metrics
UserAgent should be in the format of "[tool-name]/[version]"

For example:

USER_AGENT_PROPERTY_NAME = "mysql-socket-factory-connector-j-8/1.0.15";

shubha-rajan:
## Change Description

Updated the `SocketFactory` classes to set the user agent in `CoreSocketFactory` so that it is specific to the connector. Also updated the logic in `CoreSocketFactory.setApplicationName` to append to the existing user agent string if it has already been set, allowing third party tools like [spring-cloud-gcp](https://github.com/spring-cloud/spring-cloud-gcp/blob/8850186d11d9604ac1ce771a5313389306fa3350/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java#L173-L174) to add custom names.

## Checklist

## Relevant issues:

- Fixes #219


shubha-rajan:
Update: currently working for the postgres connector. This will be ready for review once I test the other connectors. 
image
image

shubha-rajan:
Confirmed that it's working for all connectors (screenshots from Stackdriver Trace)
image
image
image

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -31,8 +31,24 @@
 
   private Socket socket;
 
+  private static final String APPLICATION_NAME = "mysql-socket-factory-connector-j-5";
+
+  private static String getUserAgentString() {
+    try {
+      Properties packageInfo = new Properties();
+      packageInfo
+          .load(SocketFactory.class.getClassLoader().getResourceAsStream("project.properties"));
```
kurtisvg, line 40: The gcp-spring folks are getting the artifact version without having to read from a file ([example](https://github.com/spring-cloud/spring-cloud-gcp/blob/8850186d11d9604ac1ce771a5313389306fa3350/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java#L173-L174)). Can we do something similar here? 

Looks like there is a [`getImplementationtitle`](https://docs.oracle.com/javase/8/docs/api/java/lang/Package.html#getImplementationTitle--) that might be helpful as well.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -290,7 +296,8 @@ private static SQLAdmin createAdminApiClient(HttpRequestInitializer requestIniti
     JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
     SQLAdmin.Builder adminApiBuilder =
         new Builder(httpTransport, jsonFactory, requestInitializer)
-            .setApplicationName(getApplicationName());
+            .setApplicationName(
+                String.format("%s %s", getDefaultUserAgent(), getApplicationName()));
```
kurtisvg, line 300: What if this is null?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -333,6 +340,24 @@ private static KeyPair generateRsaKeyPair() {
     return generator.generateKeyPair();
   }
 
+  /** Sets the default string which is appended to the SQLAdmin API client User-Agent header. */
+  private static void setDefaultUserAgent(String userAgent) {
+    System.setProperty(APPLICATION_NAME_PROPERTY, userAgent);
```
kurtisvg, line 345: I don't think we need to store this in a system property - is only done for `getApplicationName()` for legacy reasons.

jsimonweb:
For the postgres connector, the /version seems to be getting logged as /null instead of the expected /1.0.17

shubha-rajan:
> For the postgres connector, the /version seems to be getting logged as /null instead of the expected /1.0.17

This might have been because of early testing. I tried using another method to get the version and it kept returning null. There should also be some that log the version correctly

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
shubha-rajan: I tried that earlier but kept getting null values for the version. From looking it up, I gathered it was because Maven jar builds don't include that information in the jar manifest by default. I added a config option to the parent pom.xml which added that metadata to the jar, but it still showed up as null when getting the value.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
shubha-rajan: so right now, `getDefaultUserAgent` defaults to "cloud-sql-java-connector" and `getApplicationName` defaults to an empty string. So there shouldn't be any cases of either of those returning null.

shubha-rajan: I'll probably get rid of the getDefaultUserAgent method  in the next commit though, since we don't need to set a system property. In that case I'll handle if the connector name is null.

jsimonweb:
Got it. LGTM. Logged format postgres-socket-factory/1.0.17-SNAPSHOT

shubha-rajan:
> 1. Can we use load `packageInfo` in as a static private field as well?
> 2. What happens if the user uses both MySQL and Postgres? Maybe we should maintain a list of UserAgent's that should be used?
> 3. I also think we may need to consider to take care with the name - maybe `addUserAgent`, to avoid conflicting with `setApplicationName`? thoughts?

I think `addUserAgent` would make sense if we're maintaining a list of user agents. So then when any of the connectors are loaded into memory, the user agent would be added to the list of user agents. I made some changes to incorporate that in the last commit. 

I do have a few questions:
1. If the user is using both MySQL and Postgres, the user agents are added as soon as the connector classes are loaded because of the static block, right? so then any call to `connect` would have both user agents, regardless of which connector is making the call. Is this what we want?
2. Since `CoreSocketFactory` (and the SQLAdminApi client) is instantiated after the first `connect` call, is it possible that a user using both MySQL and Postgres would instantiate the SQLAdminAPI before both classes are loaded into memory and both user agents are added?
3. What's the behavior we want for setApplicationName? This is used by external apps to add their own user agent. Would it make more sense for this string to be stored in a separate variable (so that multiple calls would overwrite it) or should it be appended to the `userAgents` list?

kurtisvg:
> 1. If the user is using both MySQL and Postgres, the user agents are added as soon as the connector classes are loaded because of the static block, right? so then any call to `connect` would have both user agents, regardless of which connector is making the call. Is this what we want?

Yes - raw number of API calls for each db isn't (IMO) a useful metric, but number of users is. 

> 2. Since `CoreSocketFactory` (and the SQLAdminApi client) is instantiated after the first `connect` call, is it possible that a user using both MySQL and Postgres would instantiate the SQLAdminAPI before both classes are loaded into memory and both user agents are added?

It should always set at least one, but I'm open to suggestions on improving to ensure both are entered. I'm not too terribly worried since this is an edge case. 

> 3. What's the behavior we want for setApplicationName? This is used by external apps to add their own user agent. Would it make more sense for this string to be stored in a separate variable (so that multiple calls would overwrite it) or should it be appended to the `userAgents` list?

I think leaving `setApplicationName` alone for now to preserve compatibility, and just manually appending whatever the property is on client initialization is fine.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -333,6 +350,28 @@ private static KeyPair generateRsaKeyPair() {
...
+    if (version != null) {
+      userAgents.add(artifactId + "/" + version);
+    } else {
+      userAgents.add(artifactId);
+    }
+  }
+
+  /** Returns the default string which is appended to the SQLAdmin API client User-Agent header. */
+  public static String getUserAgentString() {
+    return String.join(" ", userAgents);
```
kurtisvg, line 364: Should this add in `USER_TOKEN_PROPERTY_NAME` if set?

connector-j-5/src/main/java/com/google/cloud/sql/mysql/SocketFactory.java
```diff
@@ -31,6 +31,13 @@
 
   private Socket socket;
 
+  private static final String DEFAULT_APPLICATION_NAME = "mysql-socket-factory-connector-j-5";
+
+  static {
+    CoreSocketFactory.addUserAgent(DEFAULT_APPLICATION_NAME);
```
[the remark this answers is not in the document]
shubha-rajan: fixed

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -92,6 +92,22 @@
   private final SQLAdmin adminApi;
   private final int serverProxyPort;
 
+  private static List<String> userAgents = new ArrayList<String>();
+  private static String version;
+
+  static {
+    try {
+      Properties packageInfo = new Properties();
```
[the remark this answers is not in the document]
shubha-rajan: moved to separate function in latest commit

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -333,6 +350,28 @@ private static KeyPair generateRsaKeyPair() {
     return generator.generateKeyPair();
   }
 
+  /** Sets the default string which is appended to the SQLAdmin API client User-Agent header. */
+  public static void addUserAgent(String artifactId) {
+    if (version != null) {
+      userAgents.add(artifactId + "/" + version);
```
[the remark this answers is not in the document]
shubha-rajan: it's defaulted to "unknown" in the latest commit

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -333,6 +350,28 @@ private static KeyPair generateRsaKeyPair() {
...
+  public static String getUserAgentString() {
+    return String.join(" ", userAgents);
+  }
+
+  /** Returns the current User-Agent header set for the underlying SQLAdmin API client. */
+  public static String getApplicationName() {
+    if (coreSocketFactory != null) {
+      return coreSocketFactory.adminApi.getApplicationName();
+    }
+    return System.getProperty(USER_TOKEN_PROPERTY_NAME, "");
```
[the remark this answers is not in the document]
shubha-rajan: Should it? I wasn't sure if it should return the whole string, or just the string set in `setApplicationName`

core/src/main/resources/com.google.cloud.sql.core/project.properties
```diff
@@ -0,0 +1 @@
+version=${project.version}
```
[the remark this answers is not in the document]
shubha-rajan: fixed

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
shubha-rajan: Not sure. Right now I'm inlining combining the two strings when creating the SQLAdmin API client.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -333,6 +350,28 @@ private static KeyPair generateRsaKeyPair() {
     return generator.generateKeyPair();
   }
 
+  /** Sets the default string which is appended to the SQLAdmin API client User-Agent header. */
+  public static void addUserAgent(String artifactId) {
```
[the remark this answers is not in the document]
shubha-rajan: fixed

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
shubha-rajan: Never mind, changed this to append the application name in the latest set of commits

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -333,6 +337,40 @@ private static KeyPair generateRsaKeyPair() {
     return generator.generateKeyPair();
   }
 
+  private static String getVersion() {
+    try {
+      Properties packageInfo = new Properties();
+      packageInfo.load(CoreSocketFactory.class.getClassLoader().getResourceAsStream(
+          "com.google.cloud.sql/project.properties"));
+      return packageInfo.getProperty("version");
```
kurtisvg, line 345: nit: Should this have a default?

shubha-rajan: fixed

Issue #164, opened 2019-11-07
Investigate how to integrate w/ r2dbc driver


saturnism:
r2dbc driver can connect to PostgreSQL.  CloudSQL user opened an issue with use case to use r2dbc to connect to CloudSQL PostgreSQL.

https://github.com/spring-projects/spring-data-r2dbc/issues/192

r2dbc cannot accept a Socket since it's Netty based. However, it has been updated to be able to accept a `TrustManager` and `KeystoreManager`.

Propose that:
1. We can expose the factory for both of these objects, so that we can use out of the box r2dbc immediately.
1. Or, create our own `ConnectionFactory` so that we can wrap both drivers and configure them with SSL.
1. Investigate adding r2dbc driver specifically for CloudSQL PostgreSQL.

/cc @kurtisvg @meltsufin @elefeint

masoudparvari:
Hi @saturnism, any update or timeline on this? Publishing of r2dbc-postgresql 0.8.0.RELEASE is imminent and it will support Unix Domain Socket. So I am wondering from within a gcp app which already can connect to the gcp postgres with jdbc driver and through spring cloud gcp autoconfigure with providing "instance-connection-name" and datasource username/password, is it possible to replace jdbc with r2dbc and unix socket "/cloudsql/INSTANCE_CONNECTION_NAME/.s.PGSQL.5432." ?
Honestly this has stopped us from taking R2DBC into production.

meltsufin:
cc/ @elefeint

saturnism:
/cc @kurtisvg 

@masoudparvari  btw - which environment are you connecting from? App Engine, or somewhere else?

Thanks!

masoudparvari:
@saturnism from within a kubernetes pod (app) in GCE and google cloud postgres database resides within the same cluster

saturnism:
@masoudparvari The only work around I can think of right now is to use the Cloud SQL proxy as a side car, and expose a shared unix socket domain in a shared volume within the pod.   In which case, r2dbc driver should be able to connect to this unix socket domain file directly and no additional cert is needed.  It's not as simple as Spring Cloud GCP, but it should work. Would you be interested in giving that a try?

kurtisvg:
I don't currently have any plans for R2DBC support on the roadmap. I'm not overly familiar with this project - can anyone provide some additional context on why this is important and what benefits this has over the currently supported drivers?

[...]

PR #231, opened 2020-07-10
Add r2dbc support for postgresql and mysql

dmitry-s:
fixes https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/164


r2dbc/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>io.r2dbc</groupId>
+      <artifactId>r2dbc-spi</artifactId>
+      <version>0.8.2.RELEASE</version>
+    </dependency>
+    <dependency>
+      <groupId>io.r2dbc</groupId>
+      <artifactId>r2dbc-postgresql</artifactId>
+      <version>0.8.4.BUILD-SNAPSHOT</version>
```
kurtisvg, line 30: 1. Can we use non-snapshot releases for these? 

2. It doesn't make sense to pull these dependencies when the user isn't needing them. We should follow the existing pattern for the other artifacts and use separate artifacts for mysql/postgres.

r2dbc/src/main/java/com/google/cloud/sql/r2dbc/GcpConnectionFactoryProvider.java
```diff
@@ -0,0 +1,224 @@
...
+
+	/**
+	 * Legacy postgres driver option value.
+	 */
+	public static final String LEGACY_POSTGRESQL_DRIVER = "postgres";
+
+	/**
+	 * Option to configure the {@code proxyPort} if {@code port} is not configured.
+	 */
+	public static final Option<Object> PROXY_PORT = Option.valueOf("proxyPort");
```
kurtisvg, line 63: Not sure I follow this - we should essentially be ignoring the port and always be using 3307

r2dbc/src/main/java/com/google/cloud/sql/r2dbc/GcpConnectionFactoryProvider.java
```diff
@@ -0,0 +1,224 @@
...
+
+	@Override
+	public ConnectionFactory create(ConnectionFactoryOptions connectionFactoryOptions) {
```
kurtisvg, line 66: Given we need to split into two artifacts, I think it makes sense to move this logic into CoreSocketFactory and just have thin wrappers in each artifact.

r2dbc/pom.xml
dmitry-s: 1. The changes that we rely on only exist in snapshot versions currently. We need to wait until the next version is released.

2. Is it ok if I made both mysql and postgresql dependencies optional?

kurtisvg: I guess my inclination is to separate them for the following reasons:

1. We can scope to `provided`, and the necessary driver will be pulled by maven if the user doesn't specify it
2. It allows for more clear seperation between driver specific logic
3. It's consistent with the existing artifacts

Are there any advantages to using an optional scope instead?

dmitry-s: I agree with 2 and 3, will separate the module. 

(Not sure what you mean in 1 - AFAIK `provided` dependencies are provided by a runtime such as application server )

r2dbc-core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -0,0 +1,108 @@
...
+
+    return driver != null
+        && protocol != null
+        && driver.equals(getDriver())
+        && supportedProtocol(protocol);
+  }
+
+  @Override
+  public String getDriver() {
+    return "gcp";
```
kurtisvg, line 87: Does the driver really need to be different? Won't the driver be flexible?

r2dbc-core/src/main/java/com/google/cloud/sql/core/R2dbcConnectionUtils.java
```diff
@@ -0,0 +1,61 @@
...
+import static io.r2dbc.spi.ConnectionFactoryOptions.HOST;
+import static io.r2dbc.spi.ConnectionFactoryOptions.PORT;
+
+import io.netty.handler.ssl.SslContextBuilder;
+import io.r2dbc.spi.ConnectionFactoryOptions.Builder;
+import java.util.Properties;
+import java.util.function.Function;
+
+public class R2dbcConnectionUtils {
+  static SocketOrCustomizer createConnection(Builder optionBuilder, Properties properties) {
```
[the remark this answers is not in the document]
dmitry-s: good catch, I was going to remove this class

r2dbc-core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private static Function<SslContextBuilder, SslContextBuilder> createSslCustomizer(
+      Builder optionBuilder, Properties properties) {
+    String hostIp = CoreSocketFactory.getHostIp(properties);
+    optionBuilder.option(HOST, hostIp).option(PORT, CoreSocketFactory.getDefaultServerProxyPort());
```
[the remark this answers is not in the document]
dmitry-s: The customizer is invoke on every SSL connection attempt, so yes it will get the updated version.

r2dbc-core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
dmitry-s: This is the prefix for the driver name, and each implementation would add its custom protocol name that follows that prefix. Users would use urls like this:
`r2dbc:gcp:postgres://<user>:<password>@<connection string>/<database>`
`r2dbc:gcp:mysql://<user>:<password>@<connection string>/<database>`

r2dbc-core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -0,0 +1,108 @@
...
+  ConnectionFactory createFactory(
+      ConnectionFactoryOptions connectionFactoryOptions, Properties properties) {
+    Builder optionBuilder = createBuilder(connectionFactoryOptions);
+
+    // precompute SSL Data to avoid blocking calls during the connect phase.
+    CoreSocketFactory.getSslData(properties);
+    String socket = CoreSocketFactory.getUnixSocketArg(properties);
+
+    if (socket != null) {
+      return socketConnectionFactory(optionBuilder, socket);
```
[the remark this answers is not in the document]
dmitry-s: Could you clarify your question?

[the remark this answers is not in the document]
kurtisvg: This function returns a ConnectionFactory, presumable that is used to create functions. Presumably, this means this factory is reused for multiple connections - maybe even all connections? If so, I'm not sure having two different ConnectionFactories makes sense, since the user may have one connection that uses the socket arg and one that doesn't. Should there be a single ConnectionFactory that has a connect method that looks at socket to determine if it should return a socket or a regular connection?

r2dbc-core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -0,0 +1,108 @@
...
+
+  abstract boolean supportedProtocol(String protocol);
+
+  @Override
+  public ConnectionFactory create(ConnectionFactoryOptions connectionFactoryOptions) {
+    String connectionName = connectionFactoryOptions.getRequiredValue(HOST);
+    String protocol = connectionFactoryOptions.getRequiredValue(PROTOCOL);
+
+    Properties properties = new Properties();
+    properties.put(CoreSocketFactory.CLOUD_SQL_INSTANCE_PROPERTY, connectionName);
```
kurtisvg, line 50: Do we need to use a `Properties` object for this -- would it be better to just use `HOST`?

dmitry-s: We need to call  `CoreSocketFactory.getHostIp(properties)` later and it requires `Properties`

r2dbc-core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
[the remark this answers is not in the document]
dmitry-s: Do you mean we should cache the socket factory and return it if it exists?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -220,6 +224,40 @@ public static Socket connect(Properties props, String unixPathSuffix) throws IOE
...
+   */
+  public static String getHostIp(Properties props) {
+    final String csqlInstanceName = props.getProperty(CLOUD_SQL_INSTANCE_PROPERTY);
+    final List<String> ipTypes = listIpTypes(props.getProperty("ipTypes", DEFAULT_IP_TYPES));
+
+    return getInstance().getHostIp(csqlInstanceName, ipTypes);
+  }
+
+  private String getHostIp(String instanceName, List<String> ipTypes) {
+    CloudSqlInstance instance =
```
kurtisvg, line 254: nit: This `instances.computeIfAbsent(instanceName, k -> new CloudSqlInstance(k, adminApi, executor, localKeyPair));` is being used in a couple of different places now - would it make sense to move it into it's own function to avoid any instances accidentally getting created differently?

r2dbc-mysql/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>io.r2dbc</groupId>
+      <artifactId>r2dbc-spi</artifactId>
+      <version>0.8.2.RELEASE</version>
+    </dependency>
+    <dependency>
+      <groupId>dev.miku</groupId>
+      <artifactId>r2dbc-mysql</artifactId>
+      <version>0.8.2.BUILD-SNAPSHOT</version>
```
kurtisvg, line 30: Same here - does it have to be a SNAPSHOT?

r2dbc-core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -0,0 +1,112 @@
...
+import io.r2dbc.spi.ConnectionFactoryOptions;
+import io.r2dbc.spi.ConnectionFactoryProvider;
+import io.r2dbc.spi.Option;
+import java.util.Properties;
+import java.util.function.Function;
+
+/** {@link ConnectionFactoryProvider} for proxied access to GCP Postgres and MySQL instances. */
+public abstract class GcpConnectionFactoryProvider implements ConnectionFactoryProvider {
+
+  abstract ConnectionFactory tcpConnectonFactory(
```
kurtisvg, line 35: It would be great if we could get some brief comments on these interfaces - they make sense now, but I'm worried in a month or two I may forget their importance.

dmitry-s: done

r2dbc-postgres/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>io.r2dbc</groupId>
+      <artifactId>r2dbc-spi</artifactId>
+      <version>0.8.2.RELEASE</version>
+    </dependency>
+    <dependency>
+      <groupId>io.r2dbc</groupId>
+      <artifactId>r2dbc-postgresql</artifactId>
+      <version>0.8.5.BUILD-SNAPSHOT</version>
```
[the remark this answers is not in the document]
dmitry-s: It has to be 0.8.5 because of critical SSL tunnel bug that was fixed there. 0.8.5.RELEASE is not available yet.

r2dbc-postgres/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.sql</groupId>
+      <artifactId>r2cbc-connection-factory-core</artifactId>
+      <version>1.0.17-SNAPSHOT</version>
+    </dependency>
+  </dependencies>
+
+  <repositories>
```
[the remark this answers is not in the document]
dmitry-s: We should remove it when we switch to RELEASE version.

r2dbc-postgres/pom.xml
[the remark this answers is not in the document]
kurtisvg: Ok, I think given that we'll have to wait to release these artifacts until there is a release candidate. We can merge the PR before then, but we'll skip pushing these artifacts to maven until then.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -153,7 +162,7 @@ static ListeningScheduledExecutorService getDefaultExecutor() {
   }
 
   /** Extracts the Unix socket argument from specified properties object. If unset, returns null. */
-  private static String getUnixSocketArg(Properties props) {
+  static String getUnixSocketArg(Properties props) {
```
kurtisvg, line 165: I don't think I see this being used anywhere? Can this stay private now?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -63,7 +63,7 @@
  */
 public final class CoreSocketFactory {
   public static final String CLOUD_SQL_INSTANCE_PROPERTY = "cloudSqlInstance";
-  private static final String UNIX_SOCKET_PROPERTY = "unixSocketPath";
+  public static final String UNIX_SOCKET_PROPERTY = "unixSocketPath";
```
kurtisvg, line 66: I don't think I see this being used either - can this stay private?

r2dbc-mysql/src/test/java/com/google/cloud/sql/core/R2dbcMysqlIntegrationTests.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright 2019-2020 Google LLC
```
kurtisvg, line 2: This needs to be just 2020 (whichever year the file is originally added)

shubha-rajan:
@dmitry-s before merging, can you please resolve the dependency convergence errors? Looks like there are conflicting versions of `reactor-core` and `reactor-netty`

dmitry-s:
@shubha-rajan I fixed the `reactor-netty` issue, but unfortunately I can't fix the other one, as it is coming from `dev.miku:r2dbc-mysql`

shubha-rajan:
@dmitry-s you should be able to fix it by adding a `<dependencyManagement>` section to the pom.xml and then adding the preferred version. See #239 for an example of how to do this.

PR #239, opened 2020-08-11
chore(deps): add dependency management and maven-enforcer-plugin to pom files

### LINKED ISSUE #235: Depending on several ASM lib versions
+-com.google.cloud.sql:postgres-socket-factory:1.0.16
    +-com.github.jnr:jnr-unixsocket:0.30
      +-com.github.jnr:jnr-ffi:2.1.13
        +-org.ow2.asm:asm-commons:7.1
          +-org.ow2.asm:asm-tree:7.1

+-com.google.cloud.sql:postgres-socket-factory:1.0.16
    +-com.github.jnr:jnr-unixsocket:0.30
      +-com.github.jnr:jnr-ffi:2.1.13
        +-org.ow2.asm:asm-tree:7.1

+-com.google.cloud.sql:postgres-socket-factory:1.0.16
    +-org.ow2.asm:asm-util:8.0.1
      +-org.ow2.asm:asm-tree:8.0.1

FYI you can use maven-enforcer-plugin with <dependencyConvergence/> rule to catch such problems early.

shubha-rajan:
## Change Description

Added maven-enforcer-plugin to parent pom to check for dependency convergence errors and added`dependencyManagement` sections to all poms to use the newest required version. New dependency convergences should trigger build failures on renovateBot PRs so that we can manually go in and choose the correct version.

## Relevant issues:

- Fixes #235

### DIFF
--- a/connector-j-5/pom.xml
+++ b/connector-j-5/pom.xml
@@ -19,26 +19,6 @@
     SSL certificates manually.
   </description>
 
-  <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>com.google.guava</groupId>
-        <artifactId>guava</artifactId>
-        <version>29.0-android</version>
-      </dependency>
-      <dependency>
-        <groupId>com.google.http-client</groupId>
-        <artifactId>google-http-client</artifactId>
-        <version>1.35.0</version>
-      </dependency>
-      <dependency>
-        <groupId>com.google.http-client</groupId>
-        <artifactId>google-http-client-jackson2</artifactId>
-        <version>1.35.0</version>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
-
   <dependencies>
     <dependency>
       <groupId>mysql</groupId>
--- a/connector-j-5/pom.xml
+++ b/connector-j-5/pom.xml
@@ -19,6 +19,26 @@
     SSL certificates manually.
   </description>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.guava</groupId>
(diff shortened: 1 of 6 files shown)


connector-j-5/pom.xml
```diff
@@ -19,6 +19,26 @@
     SSL certificates manually.
   </description>
 
+  <dependencyManagement>
```
kurtisvg, line 22: Rather than repeat these 3 in each component, is there any reason not to put them all in the parent pom?

connector-j-5/pom.xml
```diff
@@ -19,6 +19,26 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>29.0-android</version>
```
kurtisvg, line 27: Are we sure that android is the version we should be using here?

shubha-rajan: That was the version that came up when running maven-enforcer-plugin and I just copied it. I could just change it to version 29.0.

connector-j-5/pom.xml
shubha-rajan: That would be a better idea. I wasn't sure if the required versions for the components would be different, but since they're all pretty much the same, I can move this to the parent pom.

shubha-rajan:
When I switched the guava version, the build stopped working. I'll try switching it back to see if that fixes it.

connector-j-5/pom.xml
kurtisvg: android might be the right call, I just want to be sure. What does the sql-admin-api use? We should likely match that.

PR #274, opened 2020-09-15
chore: configure release-please to auto-update versions

shubha-rajan:
## Change Description

This should update all of the versions in the pom files to whatever is set in `versions.txt`. We can also use different module names in `versions.txt` if eventually we decide that jdbc/r2dbc will have separate versions.

### DIFF
--- a/versions.txt
+++ b/versions.txt
@@ -1,0 +1,4 @@
+# Format:
+# module:released-version:current-version
+
+cloud-sql-java-connector:1.0.16:1.0.17-SNAPSHOT
--- a/.github/workflows/release-please.yaml
+++ b/.github/workflows/release-please.yaml
@@ -1,0 +1,27 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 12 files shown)


versions.txt
```diff
@@ -0,0 +1,4 @@
+# Format:
+# module:released-version:current-version
+
+cloud-sql-java-connector:1.0.16:1.0.17-SNAPSHOT
```
kurtisvg, line 4: So there is always a versions.txt? 

Does (or can) this update the version numbers in the file for the useragent as well?

kurtisvg (approved):

versions.txt
shubha-rajan: yeah. since we're reading the version for the user agent in from a pom.properties file

shubha-rajan: From what I understand, the `versions.txt` is the source of truth when it comes to the package versioning, and we can add comments in the pom files to sync up those versions with the version in `versions.txt`

kurtisvg: SGTM

PR #302, opened 2020-09-23
docs: add R2DBC connector info to README

### LINKED ISSUE #289: Missing documentation for r2dbc connectors
## Description
The repository is missing documentation for the newly released r2dbc connectors

## Solution
Include examples of adding the connectors to maven dependencies and forming a connection string for the `r2dbc-postgres` and `r2dbc-mysql` connectors

## Additional Context
See #285

shubha-rajan:
## Change Description

Added info on adding R2DBC connectors as dependencies and forming connection URLs. For now, I just included the info in existing sections, but we might want to think about reformatting this README.

## Relevant issues:

- Fixes #289

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ Include the following the project's `gradle.build`
 ```gradle
 compile 'com.google.cloud.sql:postgres-socket-factory:1.1.0'
 ```
+*Note: Also include the JDBC Driver for MySQL, `org.postgresql:postgresql:<LATEST-VERSION>`
 
 [//]: # ({x-version-update-end})
 
@@ -149,7 +150,7 @@ Include the following the project's `build.gradle`
 compile 'com.google.cloud.sql:cloud-sql-connector-r2dbc-mysql:1.1.0'
 ```
 
-*Note: Also include the R2DBC Driver for MySQL, `dev.miku:r2dbc-mysql:LATEST-VERSION`
+*Note: Also include the R2DBC Driver for MySQL, `dev.miku:r2dbc-mysql:<LATEST-VERSION>`
 
 #### PostgreSQL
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,8 +56,31 @@ Include the following the project's `build.gradle`
 ```gradle
 compile 'com.google.cloud.sql:mysql-socket-factory-connector-j-8:1.1.0'
 ```
+#### MySQL R2DBC
 
-#### PostgreSQL
+##### Maven
```
kurtisvg, line 61: Right now this feels a bit weaving - I have to skip over large parts if I only care about r2dbc. Would it be better to split into a JDBC section and a r2dbc section and have the "installing the library" sections beneath that? 

Also, when we move into different folders we can move these READEMEs into those folders and just link from the main README.

kurtisvg (approved):

README.md
shubha-rajan: Splitting into JDBC/R2DBC sections makes sense. I'll reformat this.

README.md
```diff
@@ -120,6 +125,91 @@ Note: The host portion of the JDBC URL is currently unused, and has no effect on
...
+```maven-pom
+    <dependency>
+      <groupId>com.google.cloud.sql</groupId>
+      <artifactId>cloud-sql-connector-r2dbc-mysql</artifactId>
+      <version>1.1.0</version>
+    </dependency>
+    <dependency>
+      <groupId>dev.miku</groupId>
+      <artifactId>r2dbc-mysql</artifactId>
+      <version>0.8.2.RELEASE</version>
```
kurtisvg, line 147: Should this say something like "LATEST"? I don't think we specify the drivers for the other libraries - do we need to here?

kurtisvg: same in a few other places.

kurtisvg (approved):
One nit then LGTM.

README.md
shubha-rajan: I thought it would be clearer if we specify that they have to include the driver as well, but maybe we take this out and just leave a note to also add the driver?

kurtisvg: I could go either way on it - if we add it for R2DBC we should add it for JDBC. 

We have the drivers specified as provided, but I _think_ maven should pull in the driver automatically if it's not provided somewhere else in the dependency chain?

shubha-rajan: I'll leave a note to include the JDBC postgres driver then. MySQL already has the table with the various driver versions so we're good there.

PR #303, opened 2020-09-24
feat: use regionalized instance ids to prevent global conflicts with sqladmin v1

shubha-rajan:
## Change Description

Update the SQLAdmin API to use an instance name with a region prefix with the format `<region>~<instance-name>` to prevent issues with conflicting instance names when sqladmin v1 is released and instance names become regionally unique instead of globally unique.

Note: these changes were tested locally using the integration tests from #301.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -69,7 +69,7 @@ class CloudSqlInstance {
   private final String projectId;
   private final String regionId;
   private final String instanceId;
-  private final String prefixedInstanceId;
+  private final String regionalizedInstanceId;
   private final ListenableFuture<KeyPair> keyPair;
 
   private final Object instanceDataGuard = new Object();
@@ -106,7 +106,7 @@ class CloudSqlInstance {
     this.projectId = matcher.group(1);
     this.regionId = matcher.group(3);
     this.instanceId = matcher.group(4);
-    this.prefixedInstanceId = String.format("%s~%s", this.regionId, this.instanceId);
+    this.regionalizedInstanceId = String.format("%s~%s", this.regionId, this.instanceId);
 
     this.apiClient = apiClient;
     this.executor = executor;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -448,9 +450,9 @@ private static Certificate createCertificate(String cert) throws CertificateExce
    * Checks for common errors that can occur when interacting with the Cloud SQL Admin API, and adds
    * additional context to help the user troubleshoot them.
    *
-   * @param ex exception thrown by the Admin API request
+   * @param ex           exception thrown by the Admin API request
```
kurtisvg, line 453: Was this intentional?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -105,6 +106,7 @@
     this.projectId = matcher.group(1);
     this.regionId = matcher.group(3);
     this.instanceId = matcher.group(4);
+    this.prefixedInstanceId = String.format("%s~%s", this.regionId, this.instanceId);
```
kurtisvg, line 109: nit: "prefixed" isn't very specific 
```suggestion
    this.regionalizedInstanceId = String.format("%s~%s", this.regionId, this.instanceId);
```

kurtisvg (changes requested):
Overall LGTM - please make sure we test in an e2e application before merging.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: No, my code formatter automatically does this (I have it set to Google Style).

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: fixed

kurtisvg (approved):

PR #406, opened 2021-02-17
chore: use release-please GitHub Action for releases

shubha-rajan:
## Change Description

* Updated to use release-please GitHub action instead of bot. 
* Added `.release/update_versions.sh` to simplify updating version from version.txt file everywhere it needs to be updated.
* Added maven-flatten plugin to inherit core socket factory dependency version from parent POM using `<revision>` property.

## Checklist


.release/update_versions.sh
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+VERSION=$(cat version.txt)
+# Update parent pom version
+mvn versions:set -DnewVersion=$VERSION
+mvn -N versions:update-child-modules
```
kurtisvg, line 19: If I understand this [SO post](https://stackoverflow.com/a/10582306/9414803) correctly:
1. If versions aren't specified in a child pom, it inherits from the parent pom
2. `mvn versions:set` seems to update the parent version in child poms.

Would it be better to remove the explicit child versions, and just run the one command?

kurtisvg (changes requested):

.release/update_versions.sh
kurtisvg: It looks like you can also use `-DgenerateBackupPoms=false` to avoid cleaning these up

shubha-rajan: Addressed both of these things in the last commit

.release/update_versions.sh
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+VERSION=$(cat version.txt)
+# Update parent pom version
+mvn versions:set -DnewVersion=$VERSION -DgenerateBackupPoms=false
+
+# Update revision property which is used to determine core dependency version
+mvn versions:set-property -Dproperty="revision" -DnewVersion=$VERSION -DgenerateBackupPoms=false
```
kurtisvg, line 21: I think maybe I didn't explain myself correctly. The link I had before mentioned that if we don't specify a version for the child pom, it'll inherit the version from the parent pom. This means we don't need to use `revision` or pom flattening, and `mvn versions:set` will update the parent version in all the poms.

kurtisvg (changes requested):

.release/update_versions.sh
shubha-rajan: So here we're using `revision` to substitute the version for where we add the core socket factory as a dependency and specify the version. Not sure if that one can be inherited.

kurtisvg: Totally missed that - in that case, can we just just add it to `dependencyManagement` in the parent pom, with `${project.version}`?

shubha-rajan: I think so

kurtisvg (approved):

PR #490, opened 2021-05-11
feat: add support for Postgres IAM authentication in JDBC and R2DBC connectors

### LINKED ISSUE #350: Add support for IAM database authentication
## Feature Description
Add support for IAM database authentication: https://cloud.google.com/sql/docs/postgres/authentication

## Additional Context

New IAM database authentication requires fetching new access token before initiating a new connection. It makes it tricky to use with connection pooling, because password cannot be set upfront.  Socket factory is called to establish a new connection, so it is well positioned to request fresh access tokens on every connection.

shubha-rajan:
## Change Description

Adds IAM Auth support. Some notes to keep in mind:
- Users need to include a nonempty string for the password when connecting. This will be ignored, but is required to bypass driver errors for null and empty passwords.
- `sslmode` needs to be disabled or the driver will hang indefinitely.

## Checklist

## Relevant issues:

- Fixes #350


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -72,6 +99,17 @@
   private final String regionalizedInstanceId;
   private final ListenableFuture<KeyPair> keyPair;
 
+  // defaultRefreshBuffer is the minimum amount of time for which a
+  // certificate must be valid to ensure the next refresh attempt has adequate
+  // time to complete.
+  private static Duration defaultRefreshBuffer = Duration.ofMinutes(5);
```
kurtisvg, line 105: nit: Can these two values be final?

core/src/main/java/com/google/cloud/sql/TokenSourceFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Factory for creating a token source from {@link Credential}s for IAM Authentication.
+ */
+public interface TokenSourceFactory {
+
+  /**
+   * Name of system property that can specify an alternative credential factory.
+   */
+  String TOKEN_SOURCE_FACTORY_PROPERTY = "cloudSql.socketFactory.TokenSourceFactory";
```
kurtisvg, line 30: We already have a credential socket factory right? Can we get a token source from the credentials?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -109,9 +150,33 @@
...
+        } catch (Exception err) {
+          throw new RuntimeException(err);
+        }
+      } else {
+        tokenSourceFactory = new DefaultTokenSourceFactory();
+      }
+      this.tokenSource = tokenSourceFactory.create();
+    } else {
+      // tokenSource shouldn't be used unless IAM Auth is enabled.
+      this.tokenSource = null;
```
kurtisvg, line 177: You may wish to consider using an `Optional` here instead of relying on `null`

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -341,6 +407,18 @@ private Certificate fetchEphemeralCertificate(KeyPair keyPair) {
...
+
+    if (enableIamAuth) {
+      try {
+        tokenSource.refresh();
+        String token = tokenSource.getAccessToken().getTokenValue();
+        request.setAccessToken(token);
+      } catch (IOException ex) {
+        throw addExceptionContext(
+            ex,
+            "An exception occurred while fetching access token:");
```
kurtisvg, line 419:
```suggestion
            "An exception occurred while fetching IAM auth token:");
```

core/src/main/java/com/google/cloud/sql/TokenSourceFactory.java
kurtisvg: I guess what I'm saying is we already have a property for setting the user credentials, can we reuse the user credentials instead of trying to ask for the token source from somewhere else?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -109,9 +150,33 @@
...
+          TokenSourceFactory.TOKEN_SOURCE_FACTORY_PROPERTY);
+      TokenSourceFactory tokenSourceFactory;
+
+      if (userTokenSourceFactoryClassName != null) {
+        try {
+          tokenSourceFactory =
+              (TokenSourceFactory)
+                  Class.forName(userTokenSourceFactoryClassName).newInstance();
+        } catch (Exception err) {
+          throw new RuntimeException(err);
```
kurtisvg, line 169: Is `newInstance()` throwing any checked exceptions? If not, do we need to catch and rewrap here?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -226,14 +243,18 @@ public static Socket connect(Properties props, String unixPathSuffix) throws IOE
...
   /**
    * Returns data that can be used to establish Cloud SQL SSL connection.
    */
+  public static SslData getSslData(String csqlInstanceName, boolean enableIamAuth) {
+    return getInstance().getCloudSqlInstance(csqlInstanceName, enableIamAuth).getSslData();
+  }
+
   public static SslData getSslData(String csqlInstanceName) {
-    return getInstance().getCloudSqlInstance(csqlInstanceName).getSslData();
+    return getSslData(csqlInstanceName, false);
```
kurtisvg, line 257: I'm not sure I understand this change - can you give me some more context?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -364,9 +442,39 @@ private Certificate fetchEphemeralCertificate(KeyPair keyPair) {
...
+        .minus(refreshBuffer);
+
+    if (timeUntilRefresh.isNegative()) {
+      // If the time until the certificate expires is less than the buffer, schedule the refresh
+      // closer to the expiration time
+      timeUntilRefresh = Duration.between(Instant.now(), certExpiration.toInstant())
+          .minus(Duration.ofSeconds(5));
+    }
+    return timeUntilRefresh.getSeconds();
+  }
```
kurtisvg, line 475: It seems like it would be better to store this data in the InstanceData object, so that it's calculated once

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -345,7 +372,10 @@ public HttpRequestInitializer create() {
       }
       if (credentials.createScopedRequired()) {
         credentials =
-            credentials.createScoped(Collections.singletonList(SQLAdminScopes.SQLSERVICE_ADMIN));
+            credentials.createScoped(Arrays.asList(
```
kurtisvg, line 375: Any reason to switch from `Collections` to `Arrays`?

kurtisvg (changes requested):

core/src/main/java/com/google/cloud/sql/TokenSourceFactory.java
shubha-rajan: So the Credential Factory returns credentials that are wrapped in a `HttpCredentialsAdapter` and there's no way to get credentials back from that, because they're private https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java#L64

We could change the Credential Factory to return a `GoogleCredentials` object and then convert that to the `HttpCredentialsAdapter` later, but I'm hesitant to change the existing interface

kurtisvg: That's a DPE owned library - can we PR a getter so we can have access to it?

shubha-rajan: https://github.com/googleapis/google-auth-library-java/pull/658

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
shubha-rajan: It's just overloading `getSSLData` so that it can be called with or without an `enableIamAuth` boolean passed in. If it's not passed in it defaults to false.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
shubha-rajan: Nope. it was mostly to move away from the singleton list to add another scope. If we're using the separate tokensource factory I think this is actually unneccessary, but when we eventually get the getter added to `HttpCredentialsAdapter`, we'll want the credentials object to have both scopes

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: `newInstance()` throws `InstantiationException` and `IllegalAccessException` and `forName()` throws  `ClassNotFoundException`

kurtisvg (approved):

meltsufin:
@shubha-rajan @kurtisvg 
>sslmode needs to be disabled or the driver will hang indefinitely.

Are there any security implications to disabling sslmode?

kurtisvg:
@meltsufin No - there are two potential layers of encryption - at the connection level, and at the database protocol level. Disabling at the database level is fine because the connection level is always persistent.

PR #506, opened 2021-05-24
fix: require TLSv1.3 when connecting using IAM authentication

shubha-rajan:
## Change Description

Require TLS v1.3 when connecting with IAM auth so that OAuth2 token is encrypted during handshake.

## Checklist

- [x] Ensure the tests and linter pass

## Relevant issues:

- Fixes b/183540928 (internal)

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -26,5 +26,6 @@ if [ -n "$KOKORO_GFILE_DIR" ]; then
 fi
 
 echo -e "******************** Running tests... ********************\n"
+echo $JAVA_HOME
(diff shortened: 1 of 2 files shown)


kurtisvg (approved):

dooman87:
Hi there,

Am I thinking right that without the support of TLS1.3 the IAM auth into Postgres won't work? Looks like AppEngine Standard Java 8 environment doesn't support TLS1.3 and I was thinking if there are any workarounds?

kurtisvg:
@dooman87 Unfortunately, IAM DB AuthN requires TLS 1.3 and there aren't any workarounds. Are you sure that GAE-Standard doesn't support TLS 1.3?

PR #561, opened 2021-07-22
fix: only replace refresh result if successful or current result is invalid

### LINKED ISSUE #550: Only replace refresh result if refresh was successful or current result is invalid
## Bug Description

Currently, the Java connector performs a refresh is always replaces the previous result with a new result. However sometimes, this result can fail and it replaces the current value. Instead, we should try to avoid replacing until it's complete. 

Currently the replacement happens here: 
https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#L369-L376

In the Cloud SQL proxy, it looks like this:
https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/main/proxy/proxy/client.go#L263-L277

In the prototype Go connector, it looks like this:
https://github.com/kurtisvg/cloud-sql-go-connector/blob/main/internal/cloudsql/instance.go#L194-L205

shubha-rajan:
## Checklist

## Relevant issues:

- Fixes #550

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -387,10 +387,11 @@ class CloudSqlInstance {
               } catch (Exception e) {
                 // this means the result was invalid
               }
-              if (instanceData == null || instanceData.getExpiration().toInstant().isBefore(Instant.now())) {
-                  // replace current if it is expired or invalid
-                  currentInstanceData = refreshFuture;
-                }
+              if (instanceData == null || instanceData.getExpiration().toInstant()
+                  .isBefore(Instant.now())) {
+                // replace current if it is expired or invalid
+                currentInstanceData = refreshFuture;
+              }
               nextInstanceData = Futures.immediateFuture(performRefresh());
             }
           }
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -381,17 +381,16 @@ class CloudSqlInstance {
             logger.log(Level.WARNING,
                 "An error occurred while performing refresh. Retrying immediately.", t);
             synchronized (instanceDataGuard) {
+              InstanceData instanceData = null;
               try {
-                InstanceData instanceData = getInstanceData();
-                Instant expiration = instanceData.getExpiration().toInstant();
-                if (expiration.isBefore(Instant.now())) {
-                  // replace current if it is expired
+                instanceData = getInstanceData();
+              } catch (Exception e) {
+                // this means the result was invalid
+              }
+              if (instanceData == null || instanceData.getExpiration().toInstant().isBefore(Instant.now())) {
+                  // replace current if it is expired or invalid
                   currentInstanceData = refreshFuture;
                 }
-              } catch (Exception e) {
-                // replace current if accessing it throws an error
-                currentInstanceData = refreshFuture;
-              }
               nextInstanceData = Futures.immediateFuture(performRefresh());
             }
           }
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -379,17 +379,23 @@ class CloudSqlInstance {
           public void onFailure(Throwable t) {
             logger.log(Level.WARNING,
                 "An error occurred while performing refresh. Retrying immediately.", t);
-            if (getInstanceData().getExpiration().toInstant().isBefore(Instant.now())) {
-              synchronized (instanceDataGuard) {
-                currentInstanceData = refreshFuture;
-                nextInstanceData = Futures.immediateFuture(performRefresh());
+            try {
+              InstanceData instanceData = getInstanceData();
+              Instant expiration = instanceData.getExpiration().toInstant();
+              if (expiration.isBefore(Instant.now())) {
+                synchronized (instanceDataGuard) {
+                  // replace current if it is expired
+                  currentInstanceData = refreshFuture;
+                }
               }
-            } else {
-              forceRefresh();
+            } catch (Exception e) {
+              synchronized (instanceDataGuard) {
+                // replace current if accessing it throws an error
+                currentInstanceData = refreshFuture;
+              }
             }
 
-
-
+            nextInstanceData = Futures.immediateFuture(performRefresh());
           }
         }, executor);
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -323,6 +320,7 @@ boolean forceRefresh() {
    * would expire.
    */
   private ListenableFuture<InstanceData> performRefresh() {
+    forcedRenewRateLimiter.acquire(1);
```
kurtisvg, line 323: nit:
```suggestion
    // To avoid unreasonable SQL Admin API usage, use a rate limit to throttle our usage. 
    forcedRenewRateLimiter.acquire(1);
```

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -364,18 +362,37 @@ boolean forceRefresh() {
...
+                      secondsUntilRefresh(),
+                      TimeUnit.SECONDS);
+            }
           }
-        },
-        executor);
+
+          public void onFailure(Throwable t) {
+            logger.log(Level.WARNING,
+                "An error occurred while performing refresh. Retrying immediately.", t);
```
kurtisvg, line 381: Can we include the error in this log, so it's at least surfaced somewhere?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -364,18 +362,37 @@ boolean forceRefresh() {
...
+            if (getInstanceData().getExpiration().toInstant().isBefore(Instant.now())) {
+              synchronized (instanceDataGuard) {
+                currentInstanceData = refreshFuture;
+                nextInstanceData = Futures.immediateFuture(performRefresh());
+              }
+            } else {
+              forceRefresh();
+            }
```
kurtisvg, line 389: I think the logic we want here is:
On error:
     if current is invalid:
          replace current
     next = new new refresh

Since we know this refresh failed, we always want to schedule a new one. We just don't want it to replace "current" if we think "current' might still be good.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -364,18 +362,37 @@ boolean forceRefresh() {
...
+                      TimeUnit.SECONDS);
+            }
           }
-        },
-        executor);
+
+          public void onFailure(Throwable t) {
+            logger.log(Level.WARNING,
+                "An error occurred while performing refresh. Retrying immediately.", t);
+            if (getInstanceData().getExpiration().toInstant().isBefore(Instant.now())) {
```
kurtisvg, line 382: We need to be careful here. There are at least two problems I see with the current code.
1. If "current" is in a bad state (e.i. error) it'll throw an (maybe unrelated) error and nothing past this will happen
2. if "current" _is_ this thread (because we've already forced refresh), this will - well I'm not sure. Block forever?

kurtisvg (changes requested):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: I thought that was what including the throwable as the third argument did.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: I don't think we'd ever get into a position where current is still pending, since this callback happens after refreshFuture has completed. The first problem is something to think about though. how can we check the validity of the instance data here, without potentially throwing errors. Would a try/catch work, where we try `getInstanceData` and if we get an error, we replace current

kurtisvg: Yea, I think I would go with a try/catch around it. Just to clarify - you are sure that the callback will trigger after the future is completed?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -364,18 +363,40 @@ boolean forceRefresh() {
...
+                InstanceData instanceData = getInstanceData();
+                Instant expiration = instanceData.getExpiration().toInstant();
+                if (expiration.isBefore(Instant.now())) {
+                  // replace current if it is expired
+                  currentInstanceData = refreshFuture;
+                }
+              } catch (Exception e) {
+                // replace current if accessing it throws an error
+                currentInstanceData = refreshFuture;
+              }
```
kurtisvg, line 394: nit:

```suggestion
              InstanceData instanceData = null;
              try {
                instanceData = getInstanceData();
              } catch (Exception e) {
                // this means the result was invalid
              }
              if (instanceData == null || instanceData.getExpiration().toInstant().isBefore(Instant.now())) {
                  // replace current if it is expired or invalid
                  currentInstanceData = refreshFuture;
                }
```

kurtisvg (approved):
LGTM, minus my one nit. Can we make sure we test it in Cloud Run for an extended period of time (~24hours?) before merging?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
kurtisvg: One thing to think about - I think we actually want the replace of `current` and `next` to happen before the future is returned. Otherwise you may introduce a race condition where we return, fail to connect, and force refresh before the callback is executed

shubha-rajan: Yeah, the callback will definitely trigger after the future is completed. I get what you're saying with the race condition, where a force refresh could be triggered before the callback, but how can we replace `current` before we know the future's result if our goal is to avoid replacing a valid `InstanceData` with an invalid one?

I think in the case where a force refresh is triggered before the callback, the worst that would happen is that we get an extra refresh attempt a minute later, since the rate limiter will delay the force refresh anyway. If the refresh attempt is successful, then we'd end up replacing a valid result with another (hopefully) valid result, or wait for another refresh attempt to fix it.

kurtisvg: To clarify, I think it's preferable if whatever the callback behavior is (replacing current or just rescheduling) before the future counts as "done" .

kurtisvg: Can you use `[thenApply`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#thenApply-java.util.function.Function-) for that?

enocom:
After running this over the weekend, here's what I saw:

image

That's one runtime exception in several days, but that didn't result in a 500. This PR seems solid.

PR #564, opened 2021-07-23
fix: remove dependency on internal sun.security.x509 classes

### LINKED ISSUE #551: Remove use of sun.security.x509 pacakgage
Some of our tests use `sun.security.x509`. These are consider 'internal' APIs, and aren't always in the JVM. We should try to remove their use. 

Some alternatives include `java.security.cert` or there appears to be a BouncyCastle API as well.

shubha-rajan:
## Change Description

Use bouncycastle instead of internal sun.security APIs to generate certs for tests. I looked into using `java.security.cert` but most examples I saw either used bouncycastle or used `sun.security` or some other internal API

## Checklist

- [x] Ensure the tests and linter pass

## Relevant issues:

- Fixes #551


enocom (commented):
LGTM.

kurtisvg (approved):

Issue #632, opened 2021-10-14
Undeclared dependency to google-http-client-gson


jmcomets:
The upgrade from `1.0.11` to `1.3.4` is not working for us. We're running into the following error:

```
Caused by: java.lang.NoClassDefFoundError: com/google/api/client/json/gson/GsonFactory

    at org.postgresql.util.ObjectFactory.instantiate (ObjectFactory.java:43)
    at org.postgresql.core.SocketFactoryFactory.getSocketFactory (SocketFactoryFactory.java:37)
    at org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl (ConnectionFactoryImpl.java:122)
    at org.postgresql.core.ConnectionFactory.openConnection (ConnectionFactory.java:49)
    at org.postgresql.jdbc.PgConnection.<init> (PgConnection.java:195)
    at org.postgresql.Driver.makeConnection (Driver.java:454)
    at org.postgresql.Driver.connect (Driver.java:256)
```

This seems to be due to an undeclared dependency to `google-http-client-gson`, which was introduced in [this commit](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/commit/2de5b9e3e3c4b31482462be28926d787c4a6cff8).

I'm not sure if there's anything else that depends on Jackson, but the dependency should probably be changed. I'll try to write a PR when I have a chance.

jmcomets:
Workaround:

```diff
     <dependency>
       <artifactId>postgres-socket-factory</artifactId>
       <version>1.3.4</version>
       <scope>runtime</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.http-client</groupId>
+          <artifactId>google-http-client-jackson2</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client-gson</artifactId>
+      <version>1.40.0</version>
+      <scope>runtime</scope>
     </dependency>
```

enocom:
Sorry for the slow response here. I have a fix that cleans up the dependency graph generally in addition to fixing this problem. Just working through some last minor details.

PR #634, opened 2021-10-14
fix: Ensure all required dependencies are declared

### LINKED ISSUE #656: Ensure all required dependencies are declared
In the past, we've had undeclared deps, unused declared deps, etc.

Let's add the [Maven Dependency Plugin](http://maven.apache.org/plugins/maven-dependency-plugin/) as part of the build to ensure we check for such errors as part of a PR.

See:
- https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/632
- https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/466
- https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/477
- https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/507

enocom:
Using maven-dependency-plugin, this commit cleans up our dependency
tree. The plugin is now run as part of the lint action and will catch
any undeclared used dependencies, or any declared unused dependencies.

All transitive dependencies are now declared where they are required.
The top level POM is just for build information now.

Fixes #656.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,8 +14,8 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -->
-
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.google.cloud.sql</groupId>
@@ -35,7 +35,7 @@
   <licenses>
     <license>
       <name>The Apache License, Version 2.0</name>
-      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
     </license>
   </licenses>
 
@@ -65,7 +65,7 @@
   <developers>
     <developer>
       <organization>Google</organization>
-      <organizationUrl>http://www.google.com</organizationUrl>
+      <organizationUrl>https://www.google.com</organizationUrl>
     </developer>
   </developers>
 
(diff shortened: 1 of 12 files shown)


enocom:
Related to https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/632.

enocom:
@shubha-rajan and @kurtisvg I've fixed all the dependency issues and consolidate all linting checks into a single profile.

There are a couple of questions remaining. See comments inline.

enocom:
There are some minor things to clean up. I'll tag people when it's ready for review again.

enocom:
@kurtisvg Please take another look.

After spending some time with this, I recommend we keep `dependencyManagement` at the top level pom.

kurtisvg:
> After spending some time with this, I recommend we keep `dependencyManagement` at the top level pom.

Any particular reasoning?

pom.xml
```diff
@@ -149,59 +176,61 @@
...
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>1.1.3</version>
+      <scope>test</scope>
+    </dependency>
```
kurtisvg, line 233: Why do we want these in the top level pom?

pom.xml
```diff
@@ -103,20 +105,40 @@
...
       <dependency>
         <groupId>io.projectreactor</groupId>
         <artifactId>reactor-core</artifactId>
         <version>3.4.12</version>
       </dependency>
       <dependency>
-        <groupId>io.projectreactor.netty</groupId>
-        <artifactId>reactor-netty</artifactId>
-        <version>1.0.11</version>
+        <groupId>io.netty</groupId>
```
kurtisvg, line 114: I'd like to understand more about why we are choosing to put this at the global level. It seems like they are only used in the r2dbc folder, wouldn't they be better off there?

kurtisvg (changes requested):

pom.xml
enocom: I fixed this. All deps are next to their usage now.

pom.xml
enocom: As discussed, I moved these to where they're used.

enocom:
I need to verify that the various versions match whatever renovate has already upgraded on main. Otherwise this should be good to go.

kurtisvg (approved):

PR #648, opened 2021-11-01
fix: add undeclared dependency

### LINKED ISSUE #632: Undeclared dependency to google-http-client-gson

enocom:
Fixes #632.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -15,9 +15,9 @@
  limitations under the License.
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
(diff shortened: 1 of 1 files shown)


enocom:
This  a quick fix so that we don't have to rush https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/634.

core/pom.xml
```diff
@@ -15,9 +15,9 @@
  limitations under the License.
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
enocom, line 18: This is what IntelliJ suggests when doing auto-format.

kurtisvg: 🤷

kurtisvg (approved):

PR #661, opened 2021-11-22
Added a new driver property called saFile

### LINKED ISSUE #660: SQL Driver should accept a service account (file) as property and not only as environment variable
## Feature Description
We encountered a severe problem with our Java application built on the GCP java client libraries, that connects from on-prem to several cloud SQL postgres instances. 

We have the technical requirement that we need to connect to several Cloud SQL instances with different service accounts within one instance of our JVM application. This is impossible right now because:

* Service account key files can be only set as environment variables
* The JDBC core_socket_factory only supports one service account per instance [see 1, line 90]

We require to include these features to the client library to push forward our project, ideally as soon as possible as we are blocked at the moment.

## Alternatives Considered
There is no real workaround for this issue. 

## Additional Context
Google support:  Case 29094848

choss:
## Change Description

It is now possible to set a service account with a driver property called saFile.

## Checklist

## Relevant issues:

- Fixes #660

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -43,6 +43,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
@@ -87,7 +88,7 @@ public final class CoreSocketFactory {
   private static final int DEFAULT_SERVER_PROXY_PORT = 3307;
   private static final int RSA_KEY_SIZE = 2048;
 
-  private static CoreSocketFactory coreSocketFactory;
+  private static Map<String, CoreSocketFactory> coreSocketFactory = new ConcurrentHashMap<>();
 
   private final ListenableFuture<KeyPair> localKeyPair;
   private final ConcurrentHashMap<String, CloudSqlInstance> instances = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -114,47 +114,56 @@
...
+    return getInstance("DEFAULT");
+  }
+
   /**
    * Returns the {@link CoreSocketFactory} singleton.
    */
-  public static synchronized CoreSocketFactory getInstance() {
-    if (coreSocketFactory == null) {
-      logger.info("First Cloud SQL connection, generating RSA key pair.");
+  public static synchronized CoreSocketFactory getInstance(String saFile) {
```
kurtisvg, line 124: does this need to be a public interface? Can we just call it internally?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -421,20 +422,19 @@ public static void addArtifactId(String artifactId) {
...
+    return String.join(" ", userAgents) + " " + getApplicationName(saFile);
   }
 
   /**
    * Returns the current User-Agent header set for the underlying SQLAdmin API client.
    */
-  public static String getApplicationName() {
-    if (coreSocketFactory != null) {
-      return coreSocketFactory.adminApi.getApplicationName();
+  public static String getApplicationName(String saFile) {
```
kurtisvg, line 435: These are breaking changes to the interface of this class.

kurtisvg (changes requested):
As is, this is currently a breaking change and would require us to perform a major version bump. I'm also somewhat concerned about the overhead involved in creating multiple clients for each instance, or only being able to index changes based on the SA file.

It would also be a good idea to make sure we have tests to verify the behavior of this feature if we decide to merge it. 

Additionally, you'll have to sign the CLA before we can accept any PRs. There should be steps to follow in the CA check.

PR #699, opened 2022-01-06
chore: unpin analyze

### LINKED ISSUE #690: Update maven-dependency-plugin when next version is available
The current latest version (3.2.0) has a bug. When the newer version comes out, let's unpin from the last working version and update.

For details, see https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/689.

enocom:
Fixes #690


enocom:
We'll let renovate update for us and if a build fails, we'll just close the PR until the next version becomes available.

kurtisvg (approved):

PR #760, opened 2022-02-17
feat: support creds file location 

### LINKED ISSUE #660: SQL Driver should accept a service account (file) as property and not only as environment variable

t-indumathy:
## Change Description

The changes include support for fetching service account credentials from a file location instead of reading it from the environment variable. 

## Checklist

## Relevant issues:

- Fixes #660

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -70,7 +70,7 @@ public final class CoreSocketFactory {
    * Property used to set the application name for the underlying SQLAdmin client.
    *
    * @deprecated Use {@link #setApplicationName(String)} to set the application name
-   *             programmatically.
+   * programmatically.
    */
 
   @Deprecated
@@ -94,7 +94,7 @@ public final class CoreSocketFactory {
   private final ListeningScheduledExecutorService executor;
   private final CredentialFactory credentialFactory;
   private final SQLAdmin adminApi;
-  private String credsJsonFileLoc;
+  private String serviceAccountFile;
   private final int serverProxyPort;
 
   private static List<String> userAgents = new ArrayList<String>();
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -152,6 +152,42 @@ public static synchronized CoreSocketFactory getInstance() {
...
+      SQLAdmin adminApi = createAdminApiClient(credential);
+      ListeningScheduledExecutorService executor = getDefaultExecutor();
+
+      coreSocketFactory =
+              new CoreSocketFactory(
+                      executor.submit(CoreSocketFactory::generateRsaKeyPair),
+                      adminApi,
+                      credentialFactory,
+                      DEFAULT_SERVER_PROXY_PORT,
+                      executor);
```
kurtisvg, line 185: Hmm, so this is maybe problematic because it means that every instance uses the credentials specified, even if it's not the the JDBC url. Is that what you wanted? 

If so, why is this preferable over the environment variable?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -224,7 +260,8 @@ public static Socket connect(Properties props, String unixPathSuffix) throws IOE
     // Gather parameters
     final String csqlInstanceName = props.getProperty(CLOUD_SQL_INSTANCE_PROPERTY);
     final boolean enableIamAuth = Boolean.parseBoolean(props.getProperty("enableIamAuth"));
-
+    // Get Creds File Location if passed
+    final String credsJsonFileLoc = props.getProperty("credsJsonFileLoc","");
```
kurtisvg, line 264: nit: I think something like "serviceAccountFile" might be preferable

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -452,3 +497,37 @@ public static void setApplicationName(String applicationName) {
...
+ * This Factory generates the credentials based on the credentials read from file
+ */
+class CredsJsonFileCredentialFactory implements CredentialFactory {
+  private static final Logger logger = Logger.getLogger(CredsJsonFileCredentialFactory.class.getName());
+  private String credsJsonFileLocation;
+
+  @Override
+  public void getFileLoc(String credsJsonFileLoc)
+  {
+    this.credsJsonFileLocation = credsJsonFileLoc;
```
kurtisvg, line 511: This seems more like a setter than a getter?

Can we just have a constructor that sets this to avoid needing to extend the interface?

kurtisvg (changes requested):
Thanks for take the time to open a PR. I've left some comments. Please feel free to reach out if you have questions.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
t-indumathy: Had come across this use-case: if there are multiple JVMs running on the same server, and if each of them would want to use different SAs to connect to different or same cloud sql instance(s), we'll not be able to use the env var.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
t-indumathy: done

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
t-indumathy: removed the interface changes and added constructor

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
kurtisvg: In this current implementation, this is still the same: each is using the same socket-factory and thus the same credentials as whatever initialized it. 

The general use-case is a bit of an anti-pattern, as each app should have a bit more isolation - if they are all using different credentials but still have access to each other credentials there isn't much gained.

kurtisvg:
Closing as stale.

PR #903, opened 2022-07-06
chore: enable unit tests on Mac OSX

shubha-rajan:
The issue here is that only 127.0.0.1 can be resolved as a loopback address by default on Mac.  We need to explicitly enable any other addresses or else we get a "cannot bind to [address]" error


enocom (approved):

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
```diff
@@ -114,7 +114,7 @@ public class CoreSocketFactoryTest {
   private static final String SERVER_MESSAGE = "HELLO";
 
   private static final String PUBLIC_IP = "127.0.0.1";
-  private static final String PRIVATE_IP = "127.0.1.1";
+  private static final String PRIVATE_IP = "127.0.0.2";
```
kurtisvg, line 117: nit: can we add a comment that on macos, this command needs to be run to enable the tests to run?

kurtisvg (approved):

PR #904, opened 2022-07-08
fix: add back missing dependencies to JAR files

### LINKED ISSUE #881: Brief summary of what bug or error was observed
## Bug Description

It looks like the latest release is missing some dependencies. When I update my working Data Grip connection I get an error:

`Driver class 'com.google.common.base.Preconditions' not found.`

## How to reproduce

1. Follow the steps at https://www.jetbrains.com/help/datagrip/connect-to-google-cloud-sql-instances.html#configuring-connections-with-the-cloud-sql-connector-for-java
2. Connect to database

These steps worked fine with v1.5.0 of the socket factory

## Environment

1. OS type and version: Fedora Linux 36
2. Socket Factory version: 1.6.1

shubha-rajan:
All of the classes show up in the JAR after building once these were introduced. @hessjcg was able to track down the root cause to be this change:  
https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/commit/64cfee110ec54272c45fe4b0d150182c65c8c828
The `test` scope was added to fix a failing lint check:
https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/runs/6236244256?check_suite_focus=true

Fixes #881

### DIFF
--- a/jdbc/mysql-j-5/pom.xml
+++ b/jdbc/mysql-j-5/pom.xml
@@ -133,7 +133,6 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
@@ -145,7 +144,7 @@
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
       <version>1.1.3</version>
-      <scope>compile</scope>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.zaxxer</groupId>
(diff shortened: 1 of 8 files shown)


hessjcg (approved):
LGTM

jdbc/mysql-j-5/pom.xml
```diff
@@ -133,7 +133,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <scope>test</scope>
+      <scope>compile</scope>
```
kurtisvg, line 136: does it need to be compile or should we just leave it to the default (which I think is compile)?

kurtisvg (approved):

jdbc/mysql-j-5/pom.xml
shubha-rajan: I think default scope should be fine.

PR #931, opened 2022-08-01
chore: Move most configuration out of CoreSocketFactory and into CloudSqlInstance

hessjcg:
Makes configuration specific to a cloud sql instances instead of part of a global singleton.

This change is small in size because it:
- avoids changes to the public methods of CoreSocketFactory
- avoids breaking existing unit tests with lots of mocks in them

Further work needed:
- Ensure that v2 style socket factory config is properly respected
- Ensure that r2dbc works properly with socket factory configs
- Possibly add more config properties to CloudSqlInstanceKey

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -18,7 +18,12 @@ package com.google.cloud.sql.core;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.gson.GsonFactory;
 import com.google.api.services.sqladmin.SQLAdmin;
 import com.google.api.services.sqladmin.model.ConnectSettings;
 import com.google.api.services.sqladmin.model.GenerateEphemeralCertRequest;
@@ -40,6 +45,7 @@ import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
 import java.security.KeyPair;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -119,35 +131,52 @@ class CloudSqlInstance {
...
    * @param keyPair        public/private key pair used to authenticate connections
    */
   CloudSqlInstance(
-      String connectionName,
-      SQLAdmin apiClient,
-      boolean enableIamAuth,
-      CredentialFactory tokenSourceFactory,
+      CloudSqlInstanceKey key,
+      Optional<SQLAdmin> apiClientOpt,
+      Optional<CredentialFactory> credentialFactoryOpt,
```
shubha-rajan, line 144: Are these optional for testing purposes?

hessjcg: Yes. This keeps the mock tests working. Normally, the CloudSqlInstance is responsible for creating its own CredentialFactory and SqlAdmin

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -166,6 +195,72 @@ class CloudSqlInstance {
...
+    }
+    return adminApiBuilder.build();
+  }
+
+
+  /**
+   * Returns the default string which is appended to the SQLAdmin API client User-Agent header.
+   */
+  static String getUserAgents() {
+    return String.join(" ", CoreSocketFactory.userAgents);
```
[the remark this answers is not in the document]
hessjcg: I agree this could use some work. CloudSqlInstance instantiates the SqlAdmin api client, so it needs know the user agent. That's why I moved getUserAgent() here. I left CoreSocketFactory.userAgents in place to keep the change set small.

shubha-rajan (commented):
Took another look at this and I'm thinking maybe this PR is still doing a little too much? In general, small changes where each PR does one specific thing are preferred.  Also, semantic versioning nit - I think most of these changes should be labeled as fixes instead of chores, especially if they affect functionality in any way. That way they trigger a minor version bump and show up in release notes.

Maybe we can separate out these changes into steps (and file corresponding issues if they don't already exist):

1.  Introduce the `CloudSqlInstanceKey` class and replace all uses of using the instance name for instance lookup with using the key. (fixes: issue where you can't create a connection that uses IAM auth if a non-IAM Auth connection for the user already exists)
2. Move `CredentialFactory` instantiation to `CloudSqlInstance` (fixes: issue where you can't connect to two different instances using different sets of credentials)
3. Move userAgent string setter/getter to `CloudSqlInstance` (fixes: user agent string can be more specific to individual connections, instead of listing all drivers associated with the `CoreSocketFactory`)
4. Move socket initialization to `CloudSqlInstance` (not sure if this fixes any issues from the user end, but it's just cleaner, more logical code)

It might also be time to take the breakdown above and flesh that out into a design doc.

PR #936, opened 2022-08-03
feat: enable setting ipType configuration option for SQL Server connector

### LINKED ISSUE #359: Add support for specifying IP Types for SQL Server
## Feature Description
Allow users to specify whether to connect over public or private IP when using the JDBC connector for SQL Server. Since the `mssql-jdbc` driver currently only accepts one argument for the socket factory constructor, this is blocked until the driver supports passing in a Properties object to the constructor.

shubha-rajan:
fixes #359 
Using a query string format for the socketFactoryConstructorArg allows passing in more configuration arguments without breaking users of previous versions


jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
```diff
@@ -36,8 +37,16 @@ public class SocketFactory extends javax.net.SocketFactory {
    * Implements the {@link SocketFactory} constructor, which can be used to create authenticated
    * connections to a Cloud SQL instance.
    */
-  public SocketFactory(String instanceName) {
-    this.props.setProperty(CoreSocketFactory.CLOUD_SQL_INSTANCE_PROPERTY, instanceName);
+  public SocketFactory(String socketFactoryConstructorArg) {
+    String[] s = socketFactoryConstructorArg.split("\\?");
+    this.props.setProperty(CoreSocketFactory.CLOUD_SQL_INSTANCE_PROPERTY, s[0]);
+    if (s.length == 2) {
+      String[] queryParams = s[1].split("&");
```
hessjcg, line 44: URLs have funny rules about escaping characters etc. I wonder if you could use some method on java.util.URI to parse this instead of rolling your own.

jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
```diff
@@ -36,8 +37,16 @@ public class SocketFactory extends javax.net.SocketFactory {
    * Implements the {@link SocketFactory} constructor, which can be used to create authenticated
    * connections to a Cloud SQL instance.
    */
-  public SocketFactory(String instanceName) {
-    this.props.setProperty(CoreSocketFactory.CLOUD_SQL_INSTANCE_PROPERTY, instanceName);
+  public SocketFactory(String socketFactoryConstructorArg) {
+    String[] s = socketFactoryConstructorArg.split("\\?");
```
hessjcg, line 41: You may want to use String.indexOf('?') instead of String.split(). Customer may include more than one '?' characters and you don't really want to split on all '?', just the first one.

jdbc/sqlserver/src/test/java/com/google/cloud/sql/sqlserver/JdbcSqlServerUnitTests.java
```diff
@@ -0,0 +1,49 @@
...
+  @Test
+  public void checkConnectionStringWithQueryParam() {
+    String socketFactoryConstructorArg = String.format("%s?%s=%s", CONNECTION_NAME, "ipTypes", "PRIVATE");
+    SocketFactory socketFactory = new SocketFactory(socketFactoryConstructorArg);
+    assertThat(socketFactory.props.get(CoreSocketFactory.CLOUD_SQL_INSTANCE_PROPERTY)).isEqualTo(
+        CONNECTION_NAME);
+    assertThat(socketFactory.props.get("ipTypes")).isEqualTo(
+        "PRIVATE");
+  }
+
```
hessjcg, line 48: I would love to see test that try out some malicious input. 

Here are a few to get you started:
```
"proj:inst:db?"
   expected: props.size() == 0, no NPE, no IndexOutOfBoundsExeptions

"proj:inst:db?useIamAuth=true&privateIp" 
    expected: useIamAuth=="true", privateIp==""

"proj:inst:db?token=abc%20def%20xyz%2F%26%3D%20%48%49%20%53%68%75%62%61"  
    expected: token == "abc def xyz/&= HI Shuba"
```

hessjcg (commented):
Good start. Stay away from parsing URLs without a library. Also, more tests and meaner tests.

jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
shubha-rajan: Are we using this as an actual URL though? I assumed we were just adapting the syntax because it was familiar.

jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
shubha-rajan: What would be the use case for more than one `?` character? Wouldn't that be a malformed string (which we should probably catch with an error). so far the only allowed args are true/false for `enableIamAuth` and public/private for ipType.

jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
```diff
@@ -26,7 +26,8 @@
 public class SocketFactory extends javax.net.SocketFactory {
 
   private static final Logger logger = Logger.getLogger(SocketFactory.class.getName());
-  private Properties props = new Properties();
+  // props are protected, not private, so that they can be accessed from unit tests
+  protected Properties props = new Properties();
```
kurtisvg, line 30: https://guava.dev/releases/19.0/api/docs/com/google/common/annotations/VisibleForTesting.html

jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
kurtisvg: Can we use `https://docs.oracle.com/javase/8/docs/api/java/net/URI.html`?

jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
hessjcg: My feeling is that if we're going to say "like URL syntax" in the docs, then we need to support full URL syntax. Otherwise someone is going to want to pass a query parameter value with a `=` character in it and get stuck. 

After doing a bit of research, I think your code can stay mostly the same with String.split(), but use java.net.URLDecoder.decode(url, StandardCharsets.UTF_8.name()) on the parameter keys and values.

jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
shubha-rajan: > Can we use [ParseURI](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html)?


Unfortunately not - I tried it and the `:` character in the instance name causes issues, and we can't make users escape the instance name without breaking existing users (see check failures in [53b5936](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/936/commits/53b5936e5fd56f03fc0fb6f9dceeb9d2b2bf63cc))

hessjcg:
This PR is still hanging out and has gotten stale. Can we finish it or close it?

hessjcg (approved):
LGTM

PR #939, opened 2022-08-04
fix: default to using TLSv1.3

### LINKED ISSUE #938: Prefer TLS 1.3 when possible
When possible, we should prefer TLS 1.3, or at least verify that the connector already has this behavior.

shubha-rajan:
fixes #938

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -277,10 +277,10 @@ class CloudSqlInstance {
    * preferredTypes.
    *
    * @param preferredTypes Preferred instance IP types to use. Valid IP types include "Public" and
-   * "Private".
+   *                       "Private".
    * @return returns a string representing the IP address for the instance
    * @throws IllegalArgumentException If the instance has no IP addresses matching the provided
-   * preferences.
+   *                                  preferences.
    */
   String getPreferredIp(List<String> preferredTypes) {
     Map<String, String> ipAddrs = getInstanceData().getIpAddrs();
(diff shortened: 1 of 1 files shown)


enocom (approved):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -427,19 +427,22 @@ private SslData createSslData(
...
           throw new RuntimeException(
               String.format(
-                  "[%s] Unable to create a SSLContext for the Cloud SQL instance.", connectionName)
+                  "[%s] Unable to create a SSLContext for the Cloud SQL instance.",
+                  connectionName)
                   + " TLSv1.3 is not supported for your Java version and is required to connect"
                   + " using IAM authentication",
               ex);
+        } else {
+          logger.warning("TLSv1.3 is not supported for your Java version, fallback to TLSv1.2");
```
kurtisvg, line 443: Quick check - will this get logged too much? It's at the warning level and seems like it'll get called each refresh for each instance?

kurtisvg (approved):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: Maybe I should demote it to info? I think it's worth logging, but also don't want to clutter the logs.

kurtisvg: I think warning is the right level, just concerned about the frequency. 

One a refresh is maybe an acceptable amount.

Alternatively, we could move this into a helper function that only calls the warning on the first invocation, and maybe call it when the CloudSQLinstance or socketfactory is initialized.

shubha-rajan: I think once per refresh is fine.

shubha-rajan: if later we decide the logs are too cluttered, we can investigate moving it into a helper function.

PR #953, opened 2022-08-16
fix: Add Automatic-Module-Name to MANIFEST.MF for JDK9+ module compatibility

### LINKED ISSUE #952: Library is not compatible with java-9 and above (module-info.java)
From java 9, java introduced modular dependencies logic. As per the logic governs, any backward library(which don't have module-info.java support) where name considered as the package name to use with `requires` in module-info.java. But `com.google.cloud.sql:mysql-socket-factory-connector-j-8:1.6.3` contains `8` in name that make your library incompatible with java 9 and above module. Please find the error message below for the same when we try to use with intelliJ.
```
Caused by: java.lang.IllegalArgumentException: mysql.socket.factory.connector.j.8: Invalid module name: '8' is not a Java identifier
```

Probable fix:
- Add the valid `Automatic-Module-Name` parameter in MANIFEST.MF

Please find the details here: https://www.baeldung.com/java-9-modularity

1. OS type and version: Mac
2. Java SDK version: 16
2. Socket Factory version:

shubha-rajan:
## Change Description

This change allows applications that use Java 9+ modules to depend on the socketFactory without having to move to a newer Java version for compilation. See [this article](https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers) for more info on how this works.

Including a jar in the module path where `Automatic-Module-Name` isn't in the manifest will lead to the module name being derived from the jar file name which, for some of our packages, is an illegal module name because it has a number in it. Renaming the jar file will get it to work, but a better fix is updating the `Automatic-Module-Name` in the manifest.
 
For the module name, Let's go with `com.google.cloud.sql` since that's the prefix for all of our groupIDs, and nobody else at Google uses it. There was a discussion in the Guava PR about how Google packages should handle automatic module names which discusses best practices: https://github.com/google/guava/pull/2846

## Relevant issues:
- Fixes #952


kurtisvg (approved):

pom.xml
```diff
@@ -104,6 +104,19 @@
         <version>2.11.0</version>
       </plugin>
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
```
Nikku-jsr, line 109: Can we have the version here. Otherwise you would be compromised with the stability of the release.

Nikku-jsr: <version>3.2.2</version> ?

PR #999, opened 2022-09-30
fix: downscope credentials used for IAM AuthN login

### LINKED ISSUE #996: Downscope OAuth2 token included in ephemeral certificate
The token included in the ephemeral cert should have just a `https://www.googleapis.com/auth/sqlservice.login` scope attached to it.

shubha-rajan:
## Relevant issues:

- Fixes #996

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -133,7 +133,7 @@ class CloudSqlInstance {
       boolean enableIamAuth,
       CredentialFactory tokenSourceFactory,
       ListeningScheduledExecutorService executor,
-      ListenableFuture<KeyPair> keyPair) throws IOException {
+      ListenableFuture<KeyPair> keyPair)  {
 
     Matcher matcher = CONNECTION_NAME.matcher(connectionName);
     checkArgument(
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -133,7 +133,7 @@ class CloudSqlInstance {
       boolean enableIamAuth,
       CredentialFactory tokenSourceFactory,
       ListeningScheduledExecutorService executor,
-      ListenableFuture<KeyPair> keyPair) {
+      ListenableFuture<KeyPair> keyPair) throws IOException {
 
     Matcher matcher = CONNECTION_NAME.matcher(connectionName);
     checkArgument(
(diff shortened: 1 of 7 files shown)


enocom (approved):

shubha-rajan:
Looks like the Code Coverage check wants me to add a test

enocom:
One option for writing a test: extract the logic here into a method that accepts a Credentials object. Then you could test what happens when that token does not supported GoogleCredentials, what happens when the credentials are scoped, etc.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -526,7 +528,17 @@ private Certificate fetchEphemeralCertificate(KeyPair keyPair) {
     if (enableIamAuth) {
       try {
         credentials.get().refresh();
```
kurtisvg, line 530: Can we refresh after we downscope?

kurtisvg (changes requested):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: Unfortunately no,  we get a nullpointer if we do that.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -0,0 +1,73 @@
...
+  private GoogleCredentials scopedCredentials;
+
+  @Mock
+  private OAuth2Credentials oAuth2Credentials;
+
+  @Before
+  public void setup() throws IOException {
+    MockitoAnnotations.openMocks(this);
+    when(googleCredentials.createScoped(
+        "https://www.googleapis.com/auth/sqlservice.login")).thenReturn(scopedCredentials);
```
kurtisvg, line 49: Are we sure that this is testing the right thing? Since the mock isn't set, is there a possibility this is just testing that they are null?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -525,8 +527,11 @@ private Certificate fetchEphemeralCertificate(KeyPair keyPair) {
 
     if (enableIamAuth) {
       try {
-        credentials.get().refresh();
-        String token = credentials.get().getAccessToken().getTokenValue();
+        OAuth2Credentials creds = credentials.get();
+        creds.refresh();
+        GoogleCredentials downscoped = getDownscopedCredentials(creds);
+        downscoped.refresh();
```
kurtisvg, line 533: I'd like to better understand why we need to refresh twice here?

kurtisvg (changes requested):

PR #1052, opened 2022-11-16
chore: use GH Actions concurrency setting and allow renovate rebases

shubha-rajan:


shubha-rajan:
Never mind, this won't work 🙃

PR #1072, opened 2022-12-13
chore: add integration test for MySQL r2dbc IAM authentication

shubha-rajan:


enocom (commented):
The test looks good. Why are we adding this sample here instead of the Java docs samples repo?

shubha-rajan:
> The test looks good. Why are we adding this sample here instead of the Java docs samples repo?

I think we had this as a stand-in sample until we could create full R2DBC sample applications in the java-docs-samples repo. The postgres sample is here as well. I figured we could move the tags over at the same time once the r2dbc samples are completed.

enocom (approved):

PR #1079, opened 2022-12-13
chore(release): use GitHub Actions to automate release steps

shubha-rajan:
Jobs:
 - `upload_uberjars`: Uploads uberjars using Cloud Build when a new tag is pushed, generates SHAs, then updates release notes.
 - `snapshot_pr`: bumps up version to latest snapshot, updates `pom.xml` files, and generates PR
 - `update-pom-files`: pushes extra commit to Release Please PR branch to update versions when PR is created
 
 Secrets:
`RELEASE_PROJECT`: the Google Cloud Project where the CloudBuild job should run
`RELEASE_BUCKET_NAME`: The GCS bucket to upload artifacts to


jackwotherspoon (commented):
Left a few small comments. I'd recommend maybe switching the name of `cloudbuild.yaml` back to `upload_uberjars.yaml` as it is more descriptive of what the file is doing and the contents of the file?

Also I'm curious if you could maybe add a comment or update the PR to discuss the testing of this PR? Have any of the new github actions jobs been tested? Or is the plan to test after merging?

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,100 @@
...
+        workload_identity_provider: ${{ secrets.PROVIDER_NAME }}
+        service_account: ${{ secrets.SERVICE_ACCOUNT }}
+
+    - name: Set up Cloud SDK
+      uses: google-github-actions/setup-gcloud@v1
+
+    - name: Cloud Build
+      id: cloudbuild
+      env:
+        RELEASE_PROJECT: '${{ secrets.RELEASE_PROJECT }}'
```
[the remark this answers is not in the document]
shubha-rajan: it's a different project, so AFAIK we'll need to grant permissions to the WIF service account to run builds and push to GCS. I'm not sure what APIs need to be enabled on the WIF project though, since ideally everything would run on the release project.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,100 @@
...
+        BUCKET_NAME: '${{ secrets.RELEASE_BUCKET_NAME }}'
+      run: |
+        export VERSION=$(cat version.txt)
+        gcloud config set project $RELEASE_PROJECT
+        gcloud builds submit --config .release/cloudbuild.yaml \
```
jackwotherspoon, line 48:
```suggestion
        gcloud builds submit --config .release/upload_uberjars.yaml \
```

.github/workflows/post-release.yml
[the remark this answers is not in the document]
jackwotherspoon: As far as I know the default for whenever the WIF SA is used outside the WIF project is that it will use the services in the separate project but always use the API and quotas from the WIF project. For example, the tests use the Cloud SQL Admin API quota from the WIF project even though the Cloud SQL instances are in the testing project.

jackwotherspoon (changes requested):

shubha-rajan:
Thanks for the review @jackwotherspoon. As far as testing goes, usually what would happen is that I'd test this on our next release and see if everything works from there, and merge fixes as we go. I can't really think of a better way to test a trigger on release.

Upon further research, I think we might run into an issue here:
https://stackoverflow.com/questions/69063452/github-actions-on-release-created-workflow-trigger-not-working

Since our release is created by the GitHub actions bot, we might not be able to trigger the post-release workflow. 
Some alternative ways we could do this are:
1. Use a Personal Access Token for the Release-Please bot instead of the GitHub Token. Downsides: pretty hacky, might not pass security review
2. Use the `workflow-run` event as a trigger instead. Downside: need a way to check the output of the release-please job to check if there was a release, can't use event id for release
3. merge the workflows and add jobs from `post-release` to release workflow. Downside:  can't use event id for release, less parallelization
4. use `workflow-dispatch` event to manually trigger post-release workflow. Downside: adds an extra manual step to our automation.

Which of these options do you think we should go with?

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+          release-type: simple
+          package-name: cloud-sql-java-connector
+  
+  update-pom-files:
+    runs-on: ubuntu-latest
+    needs: release-please
+    if: ${{ needs.release-please.outputs.pr }}
+    steps:
+      - name: update pom.xml files
+        run: ./.release/update_versions.sh
```
jackwotherspoon, line 41: I think we probably need to checkout the code before being able to run the`./.release/update_versions.sh` script?
```suggestion
      - name: Checkout code
        uses: actions/checkout@v3
      - name: update pom.xml files
        run: ./.release/update_versions.sh
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+
+  upload_uberjars:
+    runs-on: ubuntu-latest
+    needs: release-please
+    if: ${{ needs.release-please.outputs.releases_created }}
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
+    - name: Checkout
```
jackwotherspoon, line 56: making it a bit more descriptive
```suggestion
    - name: Checkout code
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+      uses: actions/checkout@v3
+    - id: 'auth'
+      name: 'Authenticate to Google Cloud'
+      uses: 'google-github-actions/auth@v1.0.0'
```
jackwotherspoon, line 60: nit: probably can remove the single quotes here for consistency
```suggestion
    - id: auth
      name: Authenticate to Google Cloud
      uses: google-github-actions/auth@v1.0.0
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+            tests:run
+            kokoro:force-run
+            automerge
+        branch: update-version-snapshot
+        delete-branch: true
+        draft: false
+      
+
+
+
```
jackwotherspoon, line 126: nit: too many new lines at EOF

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+    - name: create pull request
+      uses: peter-evans/create-pull-request@v4
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        title: 'chore: update to latest SNAPSHOT version'
+        body: |
+          Auto-generated by [create-pull-request][1]
+          [1]: https://github.com/peter-evans/create-pull-request
+        labels: |
+            tests:run
```
jackwotherspoon, line 117: nit: label is `tests: run` with a space
```suggestion
            tests: run
```

jackwotherspoon (changes requested):
I still think there are probably pieces of this PR that we could isolate and test prior to merging? (mainly the`update-pom-files` and `snapshot-pr` jobs)  For instance we could create a test PR that runs on the `pull_request` trigger, change the version in version.txt and test that the pom and snapshot jobs work as expected by removing the `needs` dependent jobs. Do you think this is worth trying to make sure those jobs run smoothly?

shubha-rajan:
> I still think there are probably pieces of this PR that we could isolate and test prior to merging? (mainly the`update-pom-files` and `snapshot-pr` jobs) For instance we could create a test PR that runs on the `pull_request` trigger, change the version in version.txt and test that the pom and snapshot jobs work as expected by removing the `needs` dependent jobs. Do you think this is worth trying to make sure those jobs run smoothly?

Tested both of those jobs on this PR: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/1094:
`update-pom-files`: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/actions/runs/3858198003/jobs/6576451022 ,
`snapshot-pr`: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/actions/runs/3858225875/jobs/6576507419, generated PR: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/1095

jackwotherspoon:
Awesome! Thanks for testing those, LGTM. Excited for this change to smoothen the release process hopefully 🤞

PR #1082, opened 2022-12-14
fix: throw exception on invalid IAM Authn config

### LINKED ISSUE #984: Throw exception on invalid Auto IAM AuthN Config
If a database engine doesn't support Auto IAM AuthN (i.e., MySQL, SQL Server), the connector should throw an error if a user has requested the feature.

Related: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/283.

shubha-rajan:
## Relevant issues:

- Fixes #984

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -459,6 +459,16 @@ class CloudSqlInstance {
     }
   }
 
+  static void checkDatabaseCompatibility(ConnectSettings instanceMetadata, boolean iamAuth,
+      String connectionName) {
+    if (iamAuth && instanceMetadata.getDatabaseVersion().contains("SQLSERVER")) {
+      throw new IllegalArgumentException(
+          String.format(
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -482,6 +492,9 @@ private Metadata fetchMetadata() {
                     + "instance.",
                 connectionName));
       }
+      
+      checkDatabaseCompatibility(instanceMetadata, enableIamAuth, connectionName);
```
enocom, line 496: nit: just inlining this method would match the checks above.

enocom (approved):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: I did it this way because it's easier to unit test. I think it'd actually be better if we pulled out the other checks and moved all the checks into a utils class, then tested that. It would increase our unit test coverage

PR #1097, opened 2023-01-09
feat: support credentials from Google OAuth Client

dirkluijk:
## Change Description

Instead of assuming every credential is a `HttpCredentialsAdapter`, also support `Credential` and throw error otherwise.

## Checklist

- [X] Make sure to open an issue as a 
  [bug/issue](https://github.com/GoogleCloudPlatform//issues/new/choose) 
  before writing your code!  That way we can discuss the change, evaluate 
  designs, and agree on the general idea.
- [ ] Ensure the tests and linter pass
- [ ] Appropriate documentation is updated (if necessary)

## Relevant issues:

- Fixes https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/1093

Issue #1093, opened 2023-01-06
ClassCastException when using IAM with Google OAuth Client


dirkluijk:
## Bug Description

When using the Google OAuth Client with this library (a combination that is currently used in Spring Cloud GCP), the following error occurs:

```
Caused by: java.lang.ClassCastException: class com.google.api.client.googleapis.auth.oauth2.GoogleCredential cannot be cast to class com.google.auth.http.HttpCredentialsAdapter (com.google.api.client.googleapis.auth.oauth2.GoogleCredential and com.google.auth.http.HttpCredentialsAdapter are in unnamed module of loader 'app')
```

## Example code + How to reproduce

https://github.com/dirkluijk/spring-gcp-sql-issue

Just make sure to set the `encoded-key` in `application.yaml` to a valid base64-encoded key and run the application.

## Stacktrace
```
java.lang.ClassCastException: class com.google.api.client.googleapis.auth.oauth2.GoogleCredential cannot be cast to class com.google.auth.http.HttpCredentialsAdapter (com.google.api.client.googleapis.auth.oauth2.GoogleCredential and com.google.auth.http.HttpCredentialsAdapter are in unnamed module of loader org.apache.catalina.loader.ParallelWebappClassLoader @6e8f2094)
	at com.google.cloud.sql.core.CloudSqlInstance.<init>(CloudSqlInstance.java:156)
```

## Environment

1. OS type and version: Mac OS
2. Java SDK version: 17
2. Socket Factory version: 1.8.0

## Related tickets

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1346

kurtisvg:
Hey @dirkluijk -- thanks for opening an issue, and being thorough enough to provide a minimal repro. At first glance, it looks like this is actually a problem in the spring-cloud-gcp starter (which I see you've linked an issue for). Let me loop them in on this issue as well. 

My understanding is that Google Oauth2 Library is deprecated, and that google-auth-library is the [preferred replacement.](https://github.com/googleapis/google-auth-library-java).

@meltsufin -- Can you help us out here?

@enocom FYI

meltsufin:
@JoeWang1127 Can you please take a look, since you worked on this issue before but needed a repro? Thanks!

JoeWang1127:
> @JoeWang1127 Can you please take a look, since you worked on this issue before but needed a repro? Thanks!

Sure, will do.

[...]

PR #1117, opened 2023-01-15
fix: support Google OAuth Client credentials without expiration time

dirkluijk:
## Change Description

Small fix for https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/1097 (cf024eb609cb79f4e819ecf25b74f1787885616a) to also allow credentials without an expiration time.

I only tested the previous fix with my own application default credentials, but it seems some tokens don't have the expiration time as it can be `null`.

## Checklist

- [X] Make sure to open an issue as a 
- [X] Ensure the tests and linter pass
- [X] Appropriate documentation is updated (if necessary)

## Relevant issues:

Additional fix for https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/1093.


shubha-rajan (approved):

shubha-rajan (changes requested):
Looks like these are null pointer exceptions

dirkluijk:
@shubha-rajan for some reason, I cannot run the unit tests locally (had issues before, but I also remember having them running fine). I also tried `ifconfig lo0 alias 127.0.0.2 up` several times since I am running on macOS.

Anyway, I fixed the issue from the build.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -182,7 +182,9 @@ private OAuth2Credentials parseCredentials(HttpRequestInitializer source) {
       Credential credential = (Credential) source;
       AccessToken accessToken = new AccessToken(
           credential.getAccessToken(),
-          new Date(credential.getExpirationTimeMilliseconds())
+          credential.getExpirationTimeMilliseconds() != null
+              ? new Date(credential.getExpirationTimeMilliseconds())
+              : null
```
shubha-rajan, line 187: can we use an optional here instead of this ternary? 

something like:

```
Optional<Long> expTime = Optional.ofNullable(credential.getExpirationTimeMilliseconds());

AccessToken accessToken = new AccessToken(
          credential.getAccessToken(),
          expTime.isPresent()
              ? new Date(expTime.get())
              : null
          );
```

shubha-rajan (changes requested):

shubha-rajan:
> @shubha-rajan for some reason, I cannot run the unit tests locally (had issues before, but I also remember having them running fine). I also tried `ifconfig lo0 alias 127.0.0.2 up` several times since I am running on macOS.
> 
> Anyway, I fixed the issue from the build.

Thanks for fixing that! I have one more suggestion before approving.

dirkluijk:
All right, I pushed a fix, let's see.

shubha-rajan:
> All right, I pushed a fix, let's see.

Looks like tests are passing this time. Did you see my note about using an Optional?

dirkluijk:
Ah, I already dragged in jsr305 for nullability, but Optional makes more sense indeed (writing Kotlin for some time already). Will have a look at it.

dirkluijk:
Applied your feedback! 🤞

shubha-rajan (approved):

PR #1147, opened 2023-02-07
feat: improve reliability of refresh operations

shubha-rajan:


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -282,6 +245,60 @@ private static Certificate createCertificate(String cert) throws CertificateExce
     return CertificateFactory.getInstance("X.509").generateCertificate(certStream);
   }
 
+  static void checkDatabaseCompatibility(ConnectSettings instanceMetadata, boolean iamAuth,
```
enocom, line 248: Can we move this organizational cleanup into a separate PR?

enocom (approved):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
shubha-rajan: I use intelliJ to format my code and it's honestly more work to manually undo the formatting fixes we don't want. Can we just leave it for now?

shubha-rajan: Actually I could probably make another PR with just the formatting and rebase this one when that's merged.

PR #1180, opened 2023-02-24
DRAFT: Add support for service account impersonation

fzakaria:


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -255,36 +262,42 @@ public static Socket connect(Properties props, String unixPathSuffix)
...
-    return getSslData(csqlInstanceName, false);
+    return getSslData(csqlInstanceName, false, Collections.emptyList());
   }
 
   /**
    * Returns default ip address that can be used to establish Cloud SQL connection.
    */
   public static String getHostIp(String csqlInstanceName) throws IOException {
-    return getInstance().getHostIp(csqlInstanceName, listIpTypes(DEFAULT_IP_TYPES));
+    return getInstance(null).getHostIp(csqlInstanceName, listIpTypes(DEFAULT_IP_TYPES));
```
fzakaria, line 293: These methods don't take enableIAMAuth so that means service impersonation couldn't even be used.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -385,6 +398,12 @@ private static SQLAdmin createAdminApiClient(HttpRequestInitializer requestIniti
...
   private static class ApplicationDefaultCredentialFactory implements CredentialFactory {
 
+    private final Properties props;
+
+    public ApplicationDefaultCredentialFactory(Properties props) {
+      this.props = props;
+    }
+
     @Override
     public HttpRequestInitializer create() {
```
fzakaria, line 408: Need to read the property here and create ImpersonatedCredential

enocom:
Thanks for the PR @fzakaria. We're going to pick up the service account impersonation work soon and see an opportunity to introduce a more extensible, programmatic API for things like this. We'll take this PR as reference, though.

PR #1181, opened 2023-02-24
chore: remove periodic logging of connection creation

carl-mastrangelo:
This change lowers the log level for subsequent connections to fine.  The reason is that the log doesn't offer any actionability for most users.  This log gets printed pretty frequently, scrolling up the other logs from the application.  Existing users who still want to know connections being made can increase their logger verbosity.


enocom:
@carl-mastrangelo Do you think this log line is useful? Maybe we shouldn't be logging at all.

carl-mastrangelo:
Personally I don't find any use.  I tried to keep it around for other users who may want to know.  I don't *think* there is any action to take in response to it.

enocom:
How about making this PR about removing it then?

carl-mastrangelo:
@enocom updated the PR description and PR

enocom (approved):

PR #1187, opened 2023-02-28
chore: move metadata and ephemeral cert logic into SqlAdminApiFetcher class

shubha-rajan:


core/src/main/java/com/google/cloud/sql/ApplicationDefaultCredentialFactory.java
```diff
@@ -0,0 +1,46 @@
...
+package com.google.cloud.sql;
+
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.services.sqladmin.SQLAdminScopes;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.GoogleCredentials;
+import java.io.IOException;
+import java.util.Arrays;
+
+public class ApplicationDefaultCredentialFactory implements CredentialFactory {
```
enocom, line 26: Pulling this out is one PR.

core/src/main/java/com/google/cloud/sql/ApiClientFactory.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.google.cloud.sql;
+
+import com.google.api.client.googleapis.services.json.AbstractGoogleJsonClient;
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.services.sqladmin.SQLAdmin;
+import com.google.common.base.Optional;
+
+public interface ApiClientFactory {
+  AbstractGoogleJsonClient create(HttpRequestInitializer credentials);
```
enocom, line 25: Let's make this return an `SQLAdmin` type -- the abstract client isn't what we'll be working with.

core/src/main/java/com/google/cloud/sql/ApplicationDefaultCredentialFactory.java
```diff
@@ -0,0 +1,46 @@
...
+    if (credentials.createScopedRequired()) {
+      credentials =
+          credentials.createScoped(Arrays.asList(
+              SQLAdminScopes.SQLSERVICE_ADMIN,
+              SQLAdminScopes.CLOUD_PLATFORM)
+          );
+    }
+    return new HttpCredentialsAdapter(credentials);
+  }
+}
```
enocom, line 46: Update your editor to always insert a newline to avoid these errors from GitHub.

core/src/main/java/com/google/cloud/sql/SqlAdminApiClientFactory.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.services.sqladmin.SQLAdmin;
+import com.google.common.base.Optional;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.logging.Logger;
+
+public class SqlAdminApiClientFactory implements ApiClientFactory {
```
enocom, line 31: Let's make this another PR.

core/src/main/java/com/google/cloud/sql/SqlAdminApiClientFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+  private static final Logger logger = Logger.getLogger(SqlAdminApiClientFactory.class.getName());
+  // Test properties, not for end-user use. May be changed or removed without notice.
+  private static final String API_ROOT_URL_PROPERTY = "_CLOUD_SQL_API_ROOT_URL";
```
enocom, line 35: When we get there, we should make these arguments and not properties.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -119,33 +96,22 @@ class CloudSqlInstance {
...
    * @param keyPair public/private key pair used to authenticate connections
    */
   CloudSqlInstance(
       String connectionName,
-      SQLAdmin apiClient,
+      SqlAdminApiService apiService,
       boolean enableIamAuth,
       CredentialFactory tokenSourceFactory,
       ListeningScheduledExecutorService executor,
       ListenableFuture<KeyPair> keyPair) throws IOException, InterruptedException {
```
enocom, line 111: All these extracts should be one PR at a time -- we can either do them when introducing the new associated class, or introduce the class in one PR, and then use it in another.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstanceName.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package com.google.cloud.sql.core;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class CloudSqlInstanceName {
```
enocom, line 24: Ditto here as elsewhere -- this is one PR on its own.

core/src/main/java/com/google/cloud/sql/core/Metadata.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.google.cloud.sql.core;
+
+import java.security.cert.Certificate;
+import java.util.Map;
+
+/**
+ * Represents the results of @link #fetchMetadata().
+ */
+class Metadata {
```
enocom, line 25: Ditto here.

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiService.java
```diff
@@ -0,0 +1,230 @@
...
+
+  private static final String SQL_LOGIN_SCOPE = "https://www.googleapis.com/auth/sqlservice.login";
+  SQLAdmin apiClient;
+
+  public SqlAdminApiService(ApiClientFactory sqlAdminApiFactory,
+      HttpRequestInitializer credential) {
+    apiClient = (SQLAdmin) sqlAdminApiFactory.create(credential);
+  }
+
+  static void checkDatabaseCompatibility(ConnectSettings instanceMetadata, boolean iamAuth,
```
enocom, line 52: Let's avoid static functions generally as we'll always be dealing with instances in our code.

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiService.java
```diff
@@ -0,0 +1,230 @@
...
+      // This error occurs when the project doesn't have the "Cloud SQL Admin API" enabled
+      String apiLink = "https://console.cloud.google.com/apis/api/sqladmin/overview?project="
+          + instanceName.getProjectId();
+      return new RuntimeException(String.format(
+          "[%s] The Google Cloud SQL Admin API is not enabled for the project \"%s\". Please "
+              + "use the Google Developers Console to enable it: %s",
+          instanceName.getConnectionName(), instanceName.getProjectId(), apiLink), ex);
+    } else if ("notAuthorized".equals(reason)) {
+      // This error occurs if the instance doesn't exist or the account isn't authorized
+      // TODO(kvg): Add credential account name to error string.
```
enocom, line 220: Let's also strip out stale TODOs -- either by deleting them or making an issue and fixing them.

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiService.java
```diff
@@ -0,0 +1,230 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.KeyPair;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class SqlAdminApiService {
```
enocom, line 42: I'd be curious if we can create a single method that returns everything we need. Also, there's no need to expose an asynchronous API. We can expose a synchronous API and internally use threading to run the two HTTP requests concurrently.

enocom: Landing this class will also need some good integration tests that hit the real API, too.

For testing edge cases, I'm curious if the generated SQLAdmin type has any helper methods for us.

core/src/main/java/com/google/cloud/sql/ApiClientFactory.java
shubha-rajan: It's an abstract client so that we can have a stub factory that returns a stub that implements the abstract client.

enocom: We can pass a [MockHttpTransport](https://javadoc.io/static/com.google.http-client/google-http-client/1.43.0/com/google/api/client/testing/http/MockHttpTransport.html) into the `SqlAdmin` class to create the equivalent of a stub.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstanceName.java
```diff
@@ -0,0 +1,64 @@
...
+  // Unique identifier for each Cloud SQL instance in the format "PROJECT:REGION:INSTANCE"
+  // Some legacy project ids are domain-scoped (e.g. "example.com:PROJECT:REGION:INSTANCE")
+  private static final Pattern CONNECTION_NAME =
+      Pattern.compile("([^:]+(:[^:]+)?):([^:]+):([^:]+)");
+  String projectId;
```
enocom, line 29: These should all be private.

core/src/main/java/com/google/cloud/sql/ApiClientFactory.java
enocom: And see [here](https://developers.google.com/resources/api-libraries/documentation/sqladmin/v1beta4/java/latest/com/google/api/services/sqladmin/SQLAdmin.Builder.html#Builder-com.google.api.client.http.HttpTransport-com.google.api.client.json.JsonFactory-com.google.api.client.http.HttpRequestInitializer-) for the SqlAdmin builder that takes an HttpTransport.

enocom (commented):
A few quick comments.

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiFetcher.java
```diff
@@ -0,0 +1,231 @@
...
+public class SqlAdminApiFetcher {
+
+  private static final String SQL_LOGIN_SCOPE = "https://www.googleapis.com/auth/sqlservice.login";
+  SQLAdmin apiClient;
+
+  public SqlAdminApiFetcher(SQLAdmin apiClient) {
+    this.apiClient = apiClient;
+  }
+  
+  void checkDatabaseCompatibility(ConnectSettings instanceMetadata, AuthType authType,
```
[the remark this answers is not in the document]
shubha-rajan: only if we delete the tests for it.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -436,4 +372,33 @@ Socket createSslSocket(String instanceName, List<String> ipTypes)
...
+  CloudSqlInstance getCloudSqlInstance(String instanceName, boolean enableIamAuth) {
+    if (enableIamAuth) {
+      return getCloudSqlInstance(instanceName, AuthType.IAM);
+    }
+    return getCloudSqlInstance(instanceName, AuthType.PASSWORD);
+  }
+
+
+  @VisibleForTesting
+  CloudSqlInstance getCloudSqlInstance(String instanceName, AuthType authType) {
```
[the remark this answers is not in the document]
shubha-rajan: I deleted the unused method, but kept the overload where it only takes the string because I want the default to be set in only one place.

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiFetcher.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Class that encapsulates all logic for interacting with SQLAdmin API.
+ */
+public class SqlAdminApiFetcher {
+
+  private static final String SQL_LOGIN_SCOPE = "https://www.googleapis.com/auth/sqlservice.login";
+  SQLAdmin apiClient;
```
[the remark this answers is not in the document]
shubha-rajan: ✅

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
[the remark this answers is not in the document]
enocom: We're going to be deleting all these VisibleForTesting annotations and so I'd prefer that we don't expand the problem. Already I see a few uses where we pass in Auth.Password. Let's just be consistent and do that everywhere so it's one less thing to clean up later.

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiFetcher.java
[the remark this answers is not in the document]
enocom: What is the public (or package-protected) interface for the fetcher? Seems like it should be create ephmeral cert and get instance metadata? The rest should be private, shouldn't it?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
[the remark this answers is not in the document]
shubha-rajan: fixed

enocom (approved):

PR #1191, opened 2023-03-02
chore(refactor): move application default credential factory to its own class

shubha-rajan:


core/src/main/java/com/google/cloud/sql/ApplicationDefaultCredentialFactory.java
```diff
@@ -0,0 +1,46 @@
...
+    if (credentials.createScopedRequired()) {
+      credentials =
+          credentials.createScoped(Arrays.asList(
+              SQLAdminScopes.SQLSERVICE_ADMIN,
+              SQLAdminScopes.CLOUD_PLATFORM)
+          );
+    }
+    return new HttpCredentialsAdapter(credentials);
+  }
+}
```
enocom, line 46: The red circle with a minus here means the file is missing a trailing newline -- let's set up our editors to always add this so we can stop seeing these.

core/src/main/java/com/google/cloud/sql/ApplicationDefaultCredentialFactory.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public HttpRequestInitializer create() {
+    GoogleCredentials credentials;
+    try {
+      credentials = GoogleCredentials.getApplicationDefault();
+    } catch (IOException err) {
+      throw new RuntimeException(
+          "Unable to obtain credentials to communicate with the Cloud SQL API", err);
+    }
+    if (credentials.createScopedRequired()) {
```
enocom, line 37: What does this method do? Shouldn't we always be adding OAuth2 scopes? Maybe this is a quirk of the Java API?

shubha-rajan: Yeah, it's a java quirk. There's a chance we could have credentials here that are already scoped, and then the call to createScoped would fail. See 
https://cloud.google.com/java/docs/reference/google-auth-library/latest/com.google.auth.oauth2.GoogleCredentials#com_google_auth_oauth2_GoogleCredentials_createScopedRequired__

core/src/main/java/com/google/cloud/sql/ApplicationDefaultCredentialFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.api.services.sqladmin.SQLAdminScopes;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.GoogleCredentials;
+import java.io.IOException;
+import java.util.Arrays;
+
+/**
+ * This class creates a HttpRequestInitializer from Application Default Credentials.
+ */
+public class ApplicationDefaultCredentialFactory implements CredentialFactory {
```
enocom, line 29: Can we encapsulate all credential concerns into this class?

See https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L118-L134.

core/src/test/java/com/google/cloud/sql/core/StubCredentialFactory.java
```diff
@@ -0,0 +1,46 @@
...
+package com.google.cloud.sql.core;
+
+import com.google.api.client.googleapis.testing.auth.oauth2.MockGoogleCredential;
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.cloud.sql.CredentialFactory;
+
+class StubCredentialFactory implements CredentialFactory {
+
+  String accessToken;
+  long expTime;
```
enocom, line 26: nit: let's use a more descriptive name here, `expirationTimeInMilliseconds`.

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
```diff
@@ -121,8 +122,8 @@ public class CoreSocketFactoryTest {
 
   // TODO(kvg): Remove this when updating tests to use single CoreSocketFactory
   private ListeningScheduledExecutorService defaultExecutor;
-  @Mock
-  private CredentialFactory credentialFactory;
+
+  private CredentialFactory credentialFactory = new StubCredentialFactory();
```
enocom, line 126: This can be final.

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
```diff
@@ -498,14 +493,22 @@ public void supportsCustomCredentialFactoryWithNoExpirationTime()
 
   @Test
   public void doesNotSupportNonGoogleCredentialWithIAM() throws InterruptedException, IOException {
```
enocom, line 495: This IOException can be removed.

enocom (approved):

PR #1202, opened 2023-03-10
chore(refactor): separate out testing utilities into testing base class

shubha-rajan:


core/src/test/java/com/google/cloud/sql/core/CloudSqlCoreTestingBase.java
```diff
@@ -0,0 +1,317 @@
...
+        decodeBase64StripWhitespace(TestKeys.CLIENT_PRIVATE_KEY));
+    PrivateKey privateKey = keyFactory.generatePrivate(privateKeySpec);
+
+    X509EncodedKeySpec publicKeySpec = new X509EncodedKeySpec(
+        decodeBase64StripWhitespace(TestKeys.CLIENT_PUBLIC_KEY));
+    PublicKey publicKey = keyFactory.generatePublic(publicKeySpec);
+
+    clientKeyPair = Futures.immediateFuture(new KeyPair(publicKey, privateKey));
+
+    defaultExecutor = CoreSocketFactory.getDefaultExecutor();
```
enocom, line 110: Let's leave this in the child class since we'll want to override it in each test case.

enocom (approved):

PR #1204, opened 2023-03-11
chore(refactor): replace enableIamAuth boolean with AuthType enum

shubha-rajan:


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -211,15 +212,29 @@ public static Socket connect(Properties props, String unixPathSuffix)
...
+    if (enableIamAuth) {
+      return getInstance().getCloudSqlInstance(csqlInstanceName, AuthType.IAM).getSslData();
+    }
+    return getInstance().getCloudSqlInstance(csqlInstanceName, AuthType.PASSWORD).getSslData();
+  }
+
+  /**
+   * Returns data that can be used to establish Cloud SQL SSL connection.
+   */
+  public static SslData getSslData(String csqlInstanceName, AuthType authType)
```
enocom, line 235: Does this need to be public?

shubha-rajan: It does because the Connector implementation will use this instead of the method which uses the boolean parameter.

enocom (approved):

PR #1218, opened 2023-03-21
chore: prefer synchronous interfaces

enocom:
This commit makes the SqlAdminApiFetcher use a synchronous interface. In addition, the RSA key pair generation is done once at initialization time of the core socket factory.


enocom:
Going to re-open once I sort out the callback issue described above.

PR #1231, opened 2023-03-27
chore: simplify and cleanup r2dbc implementation

enocom:
- Remove brittle unit tests in favor of integration tests
- Remove VisibleForTesting usages
- Pass just connection options builder (and not options too)
- Use Supplier interface instead of Function interface for simpler
  factoring


r2dbc/core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectionFactory.java
```diff
@@ -19,67 +19,59 @@
...
+  private final Supplier<ConnectionFactoryProvider> supplier;
   private final ConnectionFactoryOptions.Builder builder;
-  private final String csqlHostName;
+  private final String hostname;
   private final String ipTypes;
 
   /** Creates an instance of ConnectionFactory that pulls and sets host ip before delegating. */
   public CloudSqlConnectionFactory(
-      Function<ConnectionFactoryOptions, ConnectionFactory> connectionFactoryFactory,
+      Supplier<ConnectionFactoryProvider> supplier,
```
shubha-rajan, line 44: Using the supplier here is a million times better, thank you!!! 🙏

shubha-rajan (approved):

PR #1240, opened 2023-03-30
chore: clean up tests

enocom:
- Move tests closer to the related object (don't integration test everything)
- Introduce CloudSqlAdminApiTestDouble for better composable and intention revealing tests
- Prefer instance methods over static methods for better extensibility
- Rename FakeSslServer to what it fakes, i.e. "FakeCloudSqlProxyServer"
- Remove StubApiFetcherFactory and instantiate objects manually to set the stage for further object graph cleanup


shubha-rajan (dismissed):

enocom:
Yes, definitely a plan to restore test coverage. If you look at the deleted tests in CoreSocketFactory most of them are test low-level concerns (when does the SQL Admi API throw an error etc). I'm going to restore those tests at a lower level and keep the CoreSocketFactory about what it does.

core/src/test/java/com/google/cloud/sql/core/StubApiFetcherFactory.java
shubha-rajan, line 1: This class may not be all that useful now, but once we introduce connectors and connectorBuilders, it'll be necessary, so maybe we shouldn't delete it

enocom: Let's introduce it when we use it. I'm not convinced we want a factory.

enocom:
I can make the changes more apparent and in smaller chunks. Will redo this.

PR #1254, opened 2023-04-07
Remove extra backslash on jdbc url examples

nicolasromer:
The correct syntax only uses two backslashes 
for example: jdbc:mysql://localhost/dbname


enocom:
@nicolasromer Feel free to propose an improvement in the docs to make it clear that the third slash is intentional. It's a little misleading at first glance, I agree.

PR #1255, opened 2023-04-10
chore: remove nested future

enocom:


jackwotherspoon (approved):
LGTM, we did this exact same change in the Python Connector way back.

PR #1271, opened 2023-04-25
ci: Update kokoro stage job to store release artifacts

hessjcg:
We need to store release artifacts in Kokoro to generate SBOMs for released software.
When reviewing, please check that:

- The build passes
- All software artifacts generated by `mvn install` are accounted for


enocom (commented):
This looks good to me.

Add @shubha-rajan to sign off with more experience in the release mechanics.

shubha-rajan (approved):

PR #1313, opened 2023-06-02
fix: log error when token is invalid

### LINKED ISSUE #1174: Cloud SQL IAM service account authentication failed for user
### Bug Description

An otherwise valid configuration on occasion will result in `Cloud SQL IAM service account authentication failed for user`.

The backend will log: `Failed to validate access token` (as opposed to access token expired).

While customers have seen this error in Cloud Run (possibly suggesting a CPU-throttling issue with the background refresh), it also appears on GKE. 

Updating to the latest version has not resolved these occasional errors.

### Example code (or command)

_No response_

### Stacktrace

_No response_

### Steps to reproduce?

1. Deploy an app that logs in with Auto IAM AuthN
2. Wait awhile
3. Observe occasional failures

### Environment

1. OS type and version: Linux Container
2. Java SDK version: ?
3. Cloud SQL Java Socket Factory version: v1.10.0

### Additional Details

_No response_

enocom:
Fixes #1174


hessjcg (approved):
LGTM

PR #1315, opened 2023-06-05
feat: Use new certificate refresh logic 

hessjcg:
This will make the certificate refresh timing in the java jdbc connector consistent with the other connectors.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -71,6 +71,8 @@ class CloudSqlInstance {
   private final RateLimiter<Object> forcedRenewRateLimiter =
       RateLimiter.burstyBuilder(2, Duration.ofSeconds(30)).build();
 
+  private final RefreshCalculator refreshCalculator = new RefreshCalculator();
+
   @GuardedBy("instanceDataGuard")
   private ListenableFuture<InstanceData> currentInstanceData;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -112,22 +114,6 @@ class CloudSqlInstance {
     }
   }
 
-  static long secondsUntilRefresh(Date expiration) {
-    Duration timeUntilExp = Duration.between(Instant.now(), expiration.toInstant());
```
enocom, line 116: This is probably better than the long-based calculation. Let's use that below.

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/RefreshCalculator.java
```diff
@@ -0,0 +1,49 @@
...
+
+  // defaultRefreshBuffer is the minimum amount of time for which a
+  // certificate must be valid to ensure the next refresh attempt has adequate
+  // time to complete.
+  private static final Duration DEFAULT_REFRESH_BUFFER = Duration.ofMinutes(4);
+
+  long calculateSecondsUntilNextRefresh(Instant now, Instant expiration) {
+    Duration timeUntilExp = Duration.between(now, expiration);
+
+    if (timeUntilExp.compareTo(Duration.ofHours(1)) < 0) {
```
enocom, line 36: <=.

If it's exactly an hour until expiration, we want the branch's behavior.

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/RefreshCalculator.java
hessjcg: I'm leaving this as < for now so that it matches the behavior in the Go connector.

enocom (approved):

PR #1316, opened 2023-06-05
chore: Make CoreSocketFactory a real singleton.

hessjcg:
Create the singleton instance of CoreSocketFactory when the class is loaded instead of a lazy instantiation.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -71,13 +71,17 @@ public final class CoreSocketFactory {
   private static final int RSA_KEY_SIZE = 2048;
   private static final List<String> userAgents = new ArrayList<>();
   private static final String version = getVersion();
-  private static CoreSocketFactory coreSocketFactory;
+  private static final CoreSocketFactory coreSocketFactory;
   private final ListenableFuture<KeyPair> localKeyPair;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -71,14 +71,18 @@ public final class CoreSocketFactory {
...
+  private static final CoreSocketFactory coreSocketFactory;
   private final ListenableFuture<KeyPair> localKeyPair;
   private final ConcurrentHashMap<String, CloudSqlInstance> instances = new ConcurrentHashMap<>();
   private final ListeningScheduledExecutorService executor;
   private final CredentialFactory credentialFactory;
   private final int serverProxyPort;
   private final SqlAdminApiFetcher adminApiService;
 
+  static {
+    coreSocketFactory = newFactory();
```
enocom, line 83: Shall we use an enum instead?

https://stackoverflow.com/questions/70689/what-is-an-efficient-way-to-implement-a-singleton-pattern-in-java

enocom: In that case, what do you see as the benefit of this change?

hessjcg: It's a step towards getting rid of the KeyPair future, which is coming in the next few PRs.

hessjcg:
Closing, will combine with a different PR.

PR #1318, opened 2023-06-06
chore: Simplify code that fetches tokens to reduce the number of futures involved.

hessjcg:


enocom (commented):
Can we leave this step for last? I like the removal of the future, but without an understanding of what kind of performance regression there might be (if any), I'd prefer to wait and see if we actually need to make this change.

hessjcg:
Closing because we are uncertain of the performance impact on startup time.

PR #1320, opened 2023-06-06
chore: Rewrite SqlAdminApiFetcher.getInstanceData() with less use of Futures

hessjcg:


hessjcg:
Closing. We determined that this change is unnecessary.

PR #1327, opened 2023-06-13
chore: Clean up existing public java api 

hessjcg:
Removes public access from unused methods.  Moves non-public classes into the core package.  This should have no impact existing implementations using the connector.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
@@ -258,7 +258,13 @@ public final class CoreSocketFactory {
     }
   }
 
-  /** Sets the default string which is appended to the SQLAdmin API client User-Agent header. */
+  /**
+   * Internal use only: Sets the default string which is appended to the SQLAdmin API client
+   * User-Agent header.
+   *
+   * <p>This is used by the specific database connector socket factory implementations to append
+   * their database name to the user agent.
+   */
   public static void addArtifactId(String artifactId) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -291,24 +263,11 @@ private static String getUserAgents() {
...
     }
-    return System.getProperty(USER_TOKEN_PROPERTY_NAME, "");
-  }
-
-  /**
-   * Sets the User-Agent header for requests made using the underlying SQLAdmin API client.
-   *
-   * @throws IllegalStateException if the SQLAdmin client has already been initialized
-   */
-  public static void setApplicationName(String applicationName) {
```
enocom, line 306: Didn't we see usage of this in a Spring artifact?

hessjcg: You are correct. I deleted this in haste. I've updated the design doc accordingly.

enocom (approved):

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
enocom: I think we should work on removing this from spring cloud too. In theory, we can support appending the user agent, but otherwise not allow wholesale replacement of it.

hessjcg: I agree. When I update the user agent behavior, this value will be appended to the UA.

PR #1329, opened 2023-06-15
chore: add AccessTokenSupplier

hessjcg:
Between version v1.13 and v1.14 of google-auth-library-oauth2, a bug was fixed that changed the behavior of how downscope tokens are refreshed. Prior to the fix, the code could downscope an oauth2 token, then the refresh the downscoped token. After the fix, the code must first refresh the original oauth2 token, then downscope it again.

In order to incorporate this new behavior into our code, we needed to encapsulate the operation of refreshing and 
downscoping a token. This change became too big for one PR, so I split the implementation into 3 PRs.

This PR introduces a new class and interface `AccessTokenSupplier` and `DefaultAccessTokenSupplier` along with extensive unit tests. This new class encapsulates the logic to refresh a token, duplicating the logic in `CloudSqlInstance`

PR #1330 refactors `CloudSqlInstance` and `SqlAdminApiFetcher` and tests to use the new `DefaultAccessTokenSupplier`, removing the duplicate refresh token code. 

PR #1331 upgrades google-auth-library-oauth2 to the latest version.

Related to issue: #1294

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/AccessTokenSupplier.java
+++ b/core/src/main/java/com/google/cloud/sql/core/AccessTokenSupplier.java
@@ -1,9 +1,34 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.google.cloud.sql.core;
 
 import com.google.auth.oauth2.AccessToken;
 import java.io.IOException;
 import java.util.Optional;
 
+/** Supplies an AccessToken to use when authenticating with the Google API. */
+@FunctionalInterface
 interface AccessTokenSupplier {
+
+  /**
+   * Returns a valid access token or Optional.empty() when no token is available.
+   *
+   * @return the access token
+   * @throws IOException when an error occurs attempting to refresh the token.
+   */
   Optional<AccessToken> get() throws IOException;
 }
--- a/core/src/main/java/com/google/cloud/sql/core/AccessTokenSupplier.java
+++ b/core/src/main/java/com/google/cloud/sql/core/AccessTokenSupplier.java
@@ -1,0 +1,9 @@
+package com.google.cloud.sql.core;
+
+import com.google.auth.oauth2.AccessToken;
+import java.io.IOException;
+import java.util.Optional;
+
+interface AccessTokenSupplier {
+  Optional<AccessToken> get() throws IOException;
+}
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
@@ -22,7 +22,6 @@ import com.google.auth.Credentials;
 import com.google.auth.http.HttpCredentialsAdapter;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
-import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.Date;
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
@@ -33,6 +33,7 @@ import java.util.Optional;
  * configured HttpRequestInitializer.
  */
 class DefaultAccessTokenSupplier implements AccessTokenSupplier {
+
   private static final String SQL_LOGIN_SCOPE = "https://www.googleapis.com/auth/sqlservice.login";
 
   private final Optional<HttpRequestInitializer> tokenSource;
@@ -45,9 +46,7 @@ class DefaultAccessTokenSupplier implements AccessTokenSupplier {
    * @param tokenSource the token source that produces auth tokens.
    */
   DefaultAccessTokenSupplier(Optional<HttpRequestInitializer> tokenSource) {
-    this.tokenSource = tokenSource;
-    retryCount = 3;
-    retryDuration = Duration.ofSeconds(3);
+    this(tokenSource, 3, Duration.ofSeconds(3));
   }
 
   /**
@@ -57,7 +56,6 @@ class DefaultAccessTokenSupplier implements AccessTokenSupplier {
    * @param retryCount the number of attempts to refresh.
    * @param retryDuration the duration to wait between attempts.
    */
-  @VisibleForTesting
   DefaultAccessTokenSupplier(
       Optional<HttpRequestInitializer> tokenSource, int retryCount, Duration retryDuration) {
     this.tokenSource = tokenSource;
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
@@ -1,0 +1,199 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.auth.Credentials;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.AccessToken;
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  /**
+   * Parses credentials out of the HttpRequestInitializer.
+   *
+   * @return the GoogleCredentials.
+   * @throws RuntimeException when the HttpRequestInitializer is unrecognized, or if it can't
+   *     produce a GoogleCredentials instance.
+   */
+  private GoogleCredentials parseCredentials() {
```
hessjcg, line 75: Moved from [CloudSqlInstance.java:110](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/183c71ab5a6068b1f69d264cd72fbd1fb4fcbc17/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#L121)

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -0,0 +1,199 @@
...
+    return Optional.ofNullable(credentials.getExpirationTimeMilliseconds()).map(Date::new);
+  }
+
+  /**
+   * Produces a credential that only has sqlservice.login scope.
+   *
+   * @param credentials the credentials
+   * @return the downscoped credentials.
+   */
+  static GoogleCredentials getDownscopedCredentials(GoogleCredentials credentials) {
```
hessjcg, line 196: Moved from [CloudSqlInstance.java:110](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/183c71ab5a6068b1f69d264cd72fbd1fb4fcbc17/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#LL110C40-L110C40)

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  /**
+   * Extracts the expiration time from a Credential.
+   *
+   * @param credentials the token
+   * @return the expiration time, if set.
+   */
+  private Optional<Date> getTokenExpirationTime(Credential credentials) {
```
hessjcg, line 186: Moved from [CloudSqlInstance.java:262](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/183c71ab5a6068b1f69d264cd72fbd1fb4fcbc17/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#L262)

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -0,0 +1,199 @@
...
+   * @throws IllegalStateException if the token cannot be used to refresh.
+   */
+  @Override
+  public Optional<AccessToken> get() throws IOException {
+    if (tokenSource.isPresent()) {
+      RetryingCallable<Optional<AccessToken>> retries =
+          new RetryingCallable<>(
+              () -> {
+                final GoogleCredentials credentials;
+                credentials = parseCredentials();
```
hessjcg, line 128: The combined result of  [CloudSqlInstance:97](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/183c71ab5a6068b1f69d264cd72fbd1fb4fcbc17/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#L97) and [SqlAdminApiFetcher.java:301](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/183c71ab5a6068b1f69d264cd72fbd1fb4fcbc17/core/src/main/java/com/google/cloud/sql/core/SqlAdminApiFetcher.java#L301) with a few important differences: 

1. A lot more error checking.
2. We only refresh the token if it is about to expire (by default `refreshIfExpired()` will make a call to refresh if the token is expired or will expire within the next 5 minutes.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -0,0 +1,199 @@
...
+    return Optional.empty();
+  }
+
+  /**
+   * Extracts the expiration time from an AccessToken.
+   *
+   * @param token the token
+   * @return the expiration time, if set.
+   */
+  static Optional<Date> getTokenExpirationTime(Optional<AccessToken> token) {
```
hessjcg, line 172: Moved from [SqlAdminApiFetcher.java:161](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/183c71ab5a6068b1f69d264cd72fbd1fb4fcbc17/core/src/main/java/com/google/cloud/sql/core/SqlAdminApiFetcher.java#L161)

hessjcg (commented):
Comments for the reviewer.

core/src/main/java/com/google/cloud/sql/core/AccessTokenSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.google.cloud.sql.core;
+
+import com.google.auth.oauth2.AccessToken;
+import java.io.IOException;
+import java.util.Optional;
+
+/** Supplies an AccessToken to use when authenticating with the Google API. */
+@FunctionalInterface
+interface AccessTokenSupplier {
```
enocom, line 25: Could we just use `Supplier<AccessToken>` instead here?

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  /**
+   * Creates an instance with configurable retry settings.
+   *
+   * @param tokenSource the token source
+   * @param retryCount the number of attempts to refresh.
+   * @param retryDuration the duration to wait between attempts.
+   */
+  @VisibleForTesting
```
enocom, line 60: We should strive to avoid this annotation. Could we just expose this constructor as the primary one and rely on constants for default configuration?

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
enocom: side note: we need to stop using Date and instead prefer Instance. If we don't do that as part of this PR, let's do it as part of the next one.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -0,0 +1,199 @@
...
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Date;
+import java.util.Optional;
+
+/**
+ * DefaultAccessTokenSupplier produces access tokens using credentials produced by this connector's
+ * configured HttpRequestInitializer.
+ */
+class DefaultAccessTokenSupplier implements AccessTokenSupplier {
```
enocom, line 35: Is there a test class missing from this PR?

hessjcg: No: core/src/test/java/com/google/cloud/sql/core/DefaultAccessTokenSupplierTest.java

core/src/main/java/com/google/cloud/sql/core/AccessTokenSupplier.java
hessjcg: No, because we `Supplier<AccessToken>` doesn't throw an exception. We need to distinguish between when `get()` throws an IOException  meaning the refresh attempt failed to make a successful request, from when it throws an unchecked exception like IllegalStateException, meaning that the token can't be refreshed due to business logic.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
hessjcg: I want the constructor used in the main code to encapsulate details like timeout/retry. I'll rewrite this so that `DefaultAccessTokenSupplier(Optional<HttpRequestInitializer> tokenSource)` delegates to `DefaultAccessTokenSupplier(
      Optional<HttpRequestInitializer> tokenSource, int retryCount, Duration retryDuration) ` and remove the annotation.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
hessjcg: Good idea. I think a separate PR is the way to go.

enocom: Let's make sure to do this -- file an issue you're not planning on just getting to it right away.

core/src/test/java/com/google/cloud/sql/core/DefaultAccessTokenSupplierTest.java
```diff
@@ -0,0 +1,402 @@
...
+
+public class DefaultAccessTokenSupplierTest {
+
+  private final Date now = new Date();
+  private final Date past = Date.from(now.toInstant().plus(-1, ChronoUnit.HOURS));
+  private final Date future = Date.from(now.toInstant().plus(1, ChronoUnit.HOURS));
+
+  private GoogleCredentials scopedCredentials;
+  private volatile int refreshCounter = 0;
+  @Mock private OAuth2Credentials oAuth2Credentials;
```
enocom, line 57: Can we avoid using this mock?

enocom (commented):
Implementation LGTM. One question about the use of mocks and then we can move ahead.

core/src/test/java/com/google/cloud/sql/core/DefaultAccessTokenSupplierTest.java
hessjcg: Removed.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
hessjcg: See #1340

enocom (approved):

PR #1330, opened 2023-06-15
chore: Use AccessTokenSupplier, remove duplicate code from CloudSqlInstance.

hessjcg:
Refactors `CloudSqlInstance` and `SqlAdminApiFetcher` and tests to use the new `DefaultAccessTokenSupplier`, removing the duplicate refresh token code. Add unit tests to
cover CloudSqlInstance behavior.

This PR follows #1329
Related to issue #1294

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -48,7 +48,7 @@ class CloudSqlInstance {
   private static final Logger logger = Logger.getLogger(CloudSqlInstance.class.getName());
 
   private final ListeningScheduledExecutorService executor;
-  private final InstanceDataSupplier apiFetcher;
+  private final InstanceDataSupplier instanceDataSupplier;
   private final AuthType authType;
   private final AccessTokenSupplier accessTokenSupplier;
   private final CloudSqlInstanceName instanceName;
@@ -70,19 +70,19 @@ class CloudSqlInstance {
    * Initializes a new Cloud SQL instance based on the given connection name.
    *
    * @param connectionName instance connection name in the format "PROJECT_ID:REGION_ID:INSTANCE_ID"
-   * @param apiFetcher Service class for interacting with the Cloud SQL Admin API
+   * @param instanceDataSupplier Service class for interacting with the Cloud SQL Admin API
    * @param executor executor used to schedule asynchronous tasks
    * @param keyPair public/private key pair used to authenticate connections
    */
   CloudSqlInstance(
       String connectionName,
-      InstanceDataSupplier apiFetcher,
+      InstanceDataSupplier instanceDataSupplier,
       AuthType authType,
       CredentialFactory tokenSourceFactory,
       ListeningScheduledExecutorService executor,
       ListenableFuture<KeyPair> keyPair) {
     this.instanceName = new CloudSqlInstanceName(connectionName);
-    this.apiFetcher = apiFetcher;
+    this.instanceDataSupplier = instanceDataSupplier;
     this.authType = authType;
     this.executor = executor;
     this.keyPair = keyPair;
@@ -180,7 +180,7 @@ class CloudSqlInstance {
 
     try {
       InstanceData data =
-          apiFetcher.getInstanceData(
+          instanceDataSupplier.getInstanceData(
               this.instanceName, this.accessTokenSupplier, this.authType, executor, keyPair);
 
       synchronized (instanceDataGuard) {
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -16,12 +16,7 @@
 
 package com.google.cloud.sql.core;
 
-import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.http.HttpRequestInitializer;
-import com.google.auth.http.HttpCredentialsAdapter;
-import com.google.auth.oauth2.AccessToken;
-import com.google.auth.oauth2.GoogleCredentials;
-import com.google.auth.oauth2.OAuth2Credentials;
 import com.google.cloud.sql.AuthType;
 import com.google.cloud.sql.CredentialFactory;
 import com.google.common.base.Throwables;
@@ -35,7 +30,6 @@ import java.io.IOException;
 import java.security.KeyPair;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -52,12 +46,12 @@ import javax.net.ssl.SSLSocket;
 class CloudSqlInstance {
 
   private static final Logger logger = Logger.getLogger(CloudSqlInstance.class.getName());
-  private static final String SQL_LOGIN_SCOPE = "https://www.googleapis.com/auth/sqlservice.login";
 
   private final ListeningScheduledExecutorService executor;
   private final SqlAdminApiFetcher apiFetcher;
   private final AuthType authType;
-  private final Optional<OAuth2Credentials> iamAuthnCredentials;
+  private final AccessTokenSupplier accessTokenSupplier;
+  private final Optional<HttpRequestInitializer> tokenSource;
   private final CloudSqlInstanceName instanceName;
   private final ListenableFuture<KeyPair> keyPair;
   private final Object instanceDataGuard = new Object();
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -96,9 +90,11 @@ class CloudSqlInstance {
 
     if (authType == AuthType.IAM) {
       HttpRequestInitializer source = tokenSourceFactory.create();
-      this.iamAuthnCredentials = Optional.of(parseCredentials(source));
+      this.tokenSource = Optional.ofNullable(source);
+      this.accessTokenSupplier = new DefaultAccessTokenSupplier(tokenSource);
     } else {
-      this.iamAuthnCredentials = Optional.empty();
+      this.tokenSource = Optional.empty();
+      this.accessTokenSupplier = new DefaultAccessTokenSupplier(Optional.empty());
```
enocom, line 97: Could we use a supplier that always returns an empty optional here?

Something like:

``` java
this.accessTokenSuppllier = () -> Optional.empty();
```

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -96,9 +90,11 @@ class CloudSqlInstance {
 
     if (authType == AuthType.IAM) {
       HttpRequestInitializer source = tokenSourceFactory.create();
-      this.iamAuthnCredentials = Optional.of(parseCredentials(source));
+      this.tokenSource = Optional.ofNullable(source);
```
enocom, line 93: This field is never read. Can we inline it?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -52,12 +46,12 @@
 class CloudSqlInstance {
 
   private static final Logger logger = Logger.getLogger(CloudSqlInstance.class.getName());
-  private static final String SQL_LOGIN_SCOPE = "https://www.googleapis.com/auth/sqlservice.login";
 
   private final ListeningScheduledExecutorService executor;
   private final SqlAdminApiFetcher apiFetcher;
   private final AuthType authType;
-  private final Optional<OAuth2Credentials> iamAuthnCredentials;
+  private final AccessTokenSupplier accessTokenSupplier;
```
enocom, line 53: Could we just make this `Supplier<AccessToken> accessTokenSupplier`?

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiFetcher.java
```diff
@@ -96,23 +93,27 @@ private String generatePublicKeyCert(KeyPair keyPair) {
...
-      OAuth2Credentials credentials,
+      AccessTokenSupplier accessTokenSupplier,
       AuthType authType,
       ListeningScheduledExecutorService executor,
       ListenableFuture<KeyPair> keyPair)
       throws ExecutionException, InterruptedException {
 
+    ListenableFuture<Optional<AccessToken>> token =
+        executor.submit(
+            () -> accessTokenSupplier != null ? accessTokenSupplier.get() : Optional.empty());
```
enocom, line 104: If we initialize this correctly, we can always do `accessTokenSupplier.get()` here, I think.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -15,52 +15,207 @@
...
+    refreshCount = 0;
+  }
+
+  @After
+  public void teardown() {
+    executorService.shutdownNow();
+  }
+
+  @Test
+  public void testCloudSqlInstance() throws Exception {
```
enocom, line 79: Could we be more verbose in the test method name? At a glance, I don't know what we're testing here.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -15,52 +15,207 @@
...
 
-import static com.google.common.truth.Truth.assertThat;
-import static org.junit.Assert.assertThrows;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-import com.google.auth.oauth2.GoogleCredentials;
-import com.google.auth.oauth2.OAuth2Credentials;
+import com.google.cloud.sql.AuthType;
```
enocom, line 20: The bulk of this behavior lives in SqlAdminApiFetcher and AccessTokenSupplier. Could we test those interfaces directly?

hessjcg: AuthType is just part of the API. We aren't testing it here.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Fixed.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Also, I removed all the mocks from `core` tests. This test class truly tests only CloudSqlInstance behavior.

core/src/main/java/com/google/cloud/sql/core/SqlAdminApiFetcher.java
```diff
@@ -94,25 +91,41 @@ private String generatePublicKeyCert(KeyPair keyPair) {
...
+   * @param accessTokenSupplier the token supplier
+   * @param authType the auth type
+   * @param executor the executor to use
+   * @param keyPair the client-side key pair
+   * @return an InstanceData
+   * @throws ExecutionException if an exception is thrown during execution.
+   * @throws InterruptedException if the executor is interrupted.
+   */
+  @Override
+  public InstanceData getInstanceData(
```
hessjcg, line 107: This has to be public since it implements an interface.

core/src/main/java/com/google/cloud/sql/core/InstanceDataSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.google.cloud.sql.core;
+
+import com.google.cloud.sql.AuthType;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningScheduledExecutorService;
+import java.security.KeyPair;
+import java.util.concurrent.ExecutionException;
+
+interface InstanceDataSupplier {
```
hessjcg, line 26: I introduced this interface so that we can test `CloudSqlInstance` independently from `SqlAdminApiFetcher` without using mocks.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -52,12 +46,11 @@
 class CloudSqlInstance {
 
   private static final Logger logger = Logger.getLogger(CloudSqlInstance.class.getName());
-  private static final String SQL_LOGIN_SCOPE = "https://www.googleapis.com/auth/sqlservice.login";
 
   private final ListeningScheduledExecutorService executor;
-  private final SqlAdminApiFetcher apiFetcher;
+  private final InstanceDataSupplier apiFetcher;
```
enocom, line 51: Let's always match field names to type names, either exactly or as a shorter version, e.g., instanceSupplier.

core/src/main/java/com/google/cloud/sql/core/InstanceDataSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+package com.google.cloud.sql.core;
+
+import com.google.cloud.sql.AuthType;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningScheduledExecutorService;
+import java.security.KeyPair;
+import java.util.concurrent.ExecutionException;
+
+/** Internal Use Only: Gets the instance data for the CloudSqlInstance from the API. */
+interface InstanceDataSupplier {
```
enocom, line 26: Nit: shall we call this "ConnectionInfo"? By comparison, `InstanceData` seems vague.

core/src/main/java/com/google/cloud/sql/core/InstanceDataSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Internal Use Only: Gets the instance data for the CloudSqlInstance from the API.
+   *
+   * @param instanceName the name of the instance
```
enocom, line 30: If we don't have a linter complaining about these, I'd suggest leaving off params. The types are clear enough and the docs don't provide much value in my opinion.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -15,52 +15,176 @@
...
+    fetcher =
+        (instanceName, accessTokenSupplier, authType, executor, keyPair) -> {
+          Thread.sleep(100);
+          refreshCount++;
+          this.data =
+              new InstanceData(null, sslData, Date.from(Instant.now().plus(1, ChronoUnit.HOURS)));
+          return data;
+        };
+
+    instance = newCloudSqlInstance();
```
enocom, line 115: Let's inline this function -- it's a little bit of noise, but removes the indirection and helps make the test clearer.

enocom: That also helps remove the need to rely on class level variables (i.e., fetcher).

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -15,52 +15,176 @@
...
+                          "PRIVATE", "10.10.10.10"),
+                      null),
+                  sslData,
+                  Date.from(Instant.now().plus(1, ChronoUnit.HOURS)));
+          return data;
+        };
+
+    // initialize instance after mocks are set up
+    instance = newCloudSqlInstance();
+    Truth.assertThat(instance.getPreferredIp(Arrays.asList("PUBLIC", "PRIVATE")))
```
enocom, line 144: Let's use just `assertThat` here.

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InstanceDataSupplier.java
hessjcg: Good idea.  This is a big refactor and going to touch another ~100 lines of code. Can we put it in a different PR?

core/src/main/java/com/google/cloud/sql/core/InstanceDataSupplier.java
hessjcg: Fixed.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Done. I got rid of most of the fields, moved them to instance variables.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -129,18 +129,33 @@ public Optional<AccessToken> get() throws IOException {
...
                   throw new IllegalStateException(
                       "Credentials were refreshed but expiration time is in the past");
                 }
+
+                // Now, attempt to downscope the credentials and refresh again
                 GoogleCredentials downscoped = getDownscopedCredentials(credentials);
-                if (downscoped.getAccessToken() == null || downscoped.getAccessToken().equals("")) {
-                  throw new IllegalStateException(
-                      "Donwscoped credentials do not have an access token");
+                if (downscoped.getAccessToken() == null
```
enocom, line 143: This is a duplication of above largely. Could we extract it into a `validateAccessToken` or similar?

Also, there is another `validateAccessToken(accessToken);` below that includes logging in each exception branch. Shall we consolidate all that validation here? It's important that we surface this error in the logs because the refresh logic might be hiding this exception. Ideally, we match the logging we have below and include a request for users to report the error on #1174.

enocom (changes requested):
One more little detail.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -129,19 +134,41 @@ public Optional<AccessToken> get() throws IOException {
...
+
+                  String errorMessage = "Access Token has length of zero";
+                  logger.warning(errorMessage);
+
+                  throw new IllegalStateException(errorMessage);
                 }
+
+                validateAccessTokenExpiration(credentials.getAccessToken());
+
+                // Now, attempt to downscope the credentials and refresh again
```
enocom, line 147: Let's add a comment here about why we're doing this again.

enocom (approved):

enocom (approved):

PR #1348, opened 2023-07-10
chore: add debug logging aroud refresh operation

enocom:
In addition to helping customers to understand failed connections, this logging will also help debug why a refresh operation failed.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -196,7 +196,6 @@ class CloudSqlInstance {
           String.format(
               "Refresh Operation: Completed refresh with new certificate expiration at %s.",
               data.getExpiration().toInstant().toString()));
-      // schedule a replacement before the SSLContext expires;
       long secondsToRefresh =
           refreshCalculator.calculateSecondsUntilNextRefresh(
               Instant.now(), data.getExpiration().toInstant());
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -158,14 +158,22 @@ String getPreferredIp(List<String> preferredTypes) {
    */
   void forceRefresh() {
     synchronized (instanceDataGuard) {
-      // If a scheduled refresh hasn't started, perform one immediately
-      if (nextInstanceData.cancel(false)) {
+      String logMessage;
+      nextInstanceData.cancel(false);
+      if (nextInstanceData.isCancelled()) {
+        logMessage =
```
hessjcg, line 164: Why not just write the log messages instead of setting a variable? For me, logging what happens as it happens is easier to read than setting a variable and logging it later.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -175,29 +183,42 @@ void forceRefresh() {
    * would expire.
    */
   private InstanceData performRefresh() throws InterruptedException, ExecutionException {
+    logger.fine("Refresh Operation: Acquiring rate limiter permit");
     // To avoid unreasonable SQL Admin API usage, use a rate limit to throttle our usage.
     forcedRenewRateLimiter.acquirePermit();
+    logger.fine("Refresh Operation: Starting...");
```
hessjcg, line 189: Change the messages to parallel before and after blocking operation: 
- "Refresh Operation: Acquiring rate limiter permit..."
- "Refresh Operation: Acquired rate limiter permit. Starting refresh..."
- "Refresh Operation: Completed refresh, new certificate expiration at %s."
- "Refresh Operation: Next operation scheduled at %s."

enocom: Done.

PR #1388, opened 2023-07-12
chore: fix mysql 5 distribution management.

hessjcg:
Update the version number on the mysql-5 distributionManagement. It does not get updated automatically by ReleasePlease.


enocom:
Shall we update the release job in that case?  I believe that's here: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/.release/update_versions.sh.

enocom (approved):

PR #1390, opened 2023-07-17
fix: remove race condition bug in refresh logic

hessjcg:
Update the logic in forceRefresh() to reduce the churn on the thread pool when the certificate refresh API calls are failing. 

New forceRefresh() logic ensures that:

- Only 1 refresh cycle may run at a time. 
- If a refresh cycle is in progress, then it will not be canceled until it succeeds.

Add new test cases to validate race conditions, deadlocks, and concurrency.

Add additional logging to help diagnose production problems with certificate refresh.

Related to #1314

Fixes #1209 
Fixes #1159

PR #1393, opened 2023-07-18
fix: Use guava rate limiter instead of dev.failsafe

hessjcg:
For simplicity, we are removing the library dev.failsafe and using a rate limiter already included in 
the Guava library instead.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -23,9 +23,9 @@ import com.google.common.base.Throwables;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningScheduledExecutorService;
+import com.google.common.util.concurrent.RateLimiter;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
-import dev.failsafe.RateLimiter;
 import java.io.IOException;
 import java.security.KeyPair;
 import java.time.Instant;
@@ -56,7 +56,7 @@ class CloudSqlInstance {
   private final ListenableFuture<KeyPair> keyPair;
   private final Object instanceDataGuard = new Object();
   // Limit forced refreshes to 1 every minute.
-  private final RateLimiter<Object> forcedRenewRateLimiter;
+  private final RateLimiter forcedRenewRateLimiter;
 
   private final RefreshCalculator refreshCalculator = new RefreshCalculator();
 
@@ -84,7 +84,7 @@ class CloudSqlInstance {
       CredentialFactory tokenSourceFactory,
       ListeningScheduledExecutorService executor,
       ListenableFuture<KeyPair> keyPair,
-      RateLimiter<Object> forcedRenewRateLimiter) {
+      RateLimiter forcedRenewRateLimiter) {
     this.instanceName = new CloudSqlInstanceName(connectionName);
     this.instanceDataSupplier = instanceDataSupplier;
     this.authType = authType;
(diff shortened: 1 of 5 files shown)


enocom:
Can we make the base branch main?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -71,7 +70,7 @@ class CloudSqlInstance {
   private boolean forceRefreshRunning;
 
   static final RateLimiter defaultRateLimiter() {
-    return RateLimiter.burstyBuilder(2, Duration.ofSeconds(30)).build();
+    return RateLimiter.create(2.0 / 60.0); // 2 requests/minute
```
enocom, line 73: We'll 1 request every 30 seconds to mostly match the other connectors.

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: 1 request every 30 seconds is exactly what this means.

enocom: Why not write 30.0 then? Also the comment is misleading.

hessjcg: The original PR from long ago had 2.0/60.0.  I have updated this PR for clarity to be 1.0/30.0.

hessjcg:
Also, regarding target branch. This PR needs to land after #1390. Once we approve and merge #1390, the base branch on this one will automatically update to be `main`.

enocom:
Looks like the base branch did change, but now we're stuck with a bunch of conflicts. In the future, we can always make main the base branch, I can approve work ahead of time and merge conflicts can be handled async.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -359,6 +358,6 @@ CloudSqlInstance getCloudSqlInstance(String instanceName, AuthType authType) {
                 credentialFactory,
                 executor,
                 localKeyPair,
-                RateLimiter.burstyBuilder(2, Duration.ofSeconds(30)).build()));
+                RateLimiter.create(1.0 / 30.0))); // 1 refresh attempt every 30 seconds
```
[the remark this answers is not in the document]
hessjcg: I don't think this makes sense. It makes no difference whether we write `RateLimiter.create(1.0/30.0)` or `RateLimiter.create(2.0/60.0)` in the code. Both statements will be compiled to `RateLimiter.create(0.03333333333)`, creating a rate limiter that allows 0.03333333333 requests per second. There is no way to configure a burst in the Guava rate limiter.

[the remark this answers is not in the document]
enocom: :facepalm: yes of course

Then I have no idea why it was `2.0 / 60.0`.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceConcurrencyTest.java
```diff
@@ -214,7 +213,7 @@ private Thread startForceRefreshThread(CloudSqlInstance inst) {
     return t;
   }
 
-  private RateLimiter<Object> newRateLimiter() {
-    return RateLimiter.burstyBuilder(2, Duration.ofMillis(50)).build();
+  private RateLimiter newRateLimiter() {
+    return RateLimiter.create(20.0); // 20/sec = every 50 ms
```
enocom, line 217: Shouldn't this match our production code?

hessjcg: No. The concurrency tests reduce length of both modeled response times and timeouts from real behavior so that the tests run faster and we get more "shots on goal" to produce a deadlock or other threading problem.

enocom (approved):

PR #1394, opened 2023-07-18
fix: Eliminate busy wait inside of the CloudSqlInstance.performRefresh().

hessjcg:
Instead of blocking until SqlAdminApiFetcher.getInstance() completes, CloudSqlInstance.performRefresh() will
listen for completion of the future and then update its internal state. This uses the threads in the
thread pool more efficiently, so that fewer threads are needed for the process to run smoothly.


enocom:
Let's talk about the design choices here first before we try to get this merged.

PR #1401, opened 2023-08-01
feat: Add support for Service Account Impersonation.

hessjcg:


core/src/main/java/com/google/cloud/sql/CredentialFactory.java
```diff
@@ -26,4 +27,8 @@ public interface CredentialFactory {
   String CREDENTIAL_FACTORY_PROPERTY = "cloudSql.socketFactory.credentialFactory";
 
   HttpRequestInitializer create();
+
+  default CredentialFactory withDelegates(List<String> delegates) {
+    return this;
```
enocom, line 32: I assume this makes the change non-breaking?

Can we write a test that uses the interface as it was, to confirm that this change doesn't break it?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -341,23 +349,31 @@ Socket createSslSocket(String instanceName, List<String> ipTypes, AuthType authT
     }
   }
 
+  @VisibleForTesting
```
enocom, line 352: What would it take to test through the public interface? Ditto below.

enocom (commented):
Some minor comments. As much as I don't love the CredentialFactory interface, we're stuck with it and this seems like the right change.

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
hessjcg: Yes, but I think there may be a better way to handle the "with delegates" issue than adding a new method to CredentialFactory. Let me try again on this one.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
hessjcg: This method is only ever used in the test cases. I will update the test cases instead and get rid of this method.

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
```diff
@@ -26,4 +32,55 @@ public interface CredentialFactory {
...
 
   HttpRequestInitializer create();
+
+  /**
+   * Returns a GoogleCredentials to use for generating tokens and authenticating web requests. The
+   * default implementation will attempt to extract the GoogleCredentials object from known
+   * credential HttpRequestInitializer implementations already in used to provide tokens for IAM
+   * Authentication.
+   */
+  default GoogleCredentials getCredentials() {
```
hessjcg, line 42: The default implementation of getCredentials() now has code that can extract a GoogleCredentials instance from well known implementations of HttpRequestInitializer. This code has been in use in `DefaultAccessTokenSupplier` for a long time to supply tokens for IAM Authentication.

core/src/main/java/com/google/cloud/sql/core/ApplicationDefaultCredentialFactory.java
```diff
@@ -29,18 +29,25 @@ class ApplicationDefaultCredentialFactory implements CredentialFactory {
 
   @Override
   public HttpRequestInitializer create() {
+    GoogleCredentials credentials = getCredentials();
+    return new HttpCredentialsAdapter(credentials);
+  }
+
+  @Override
+  public GoogleCredentials getCredentials() {
```
hessjcg, line 37: Extract existing `getCredentials()` implementation out of `create()`.

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -50,7 +48,7 @@ class DefaultAccessTokenSupplier implements AccessTokenSupplier {
    *
    * @param tokenSource the token source that produces auth tokens.
    */
-  DefaultAccessTokenSupplier(HttpRequestInitializer tokenSource) {
+  DefaultAccessTokenSupplier(CredentialFactory tokenSource) {
```
hessjcg, line 51: Replace `HttpRequestInitializer` with `CredentialFactory`. The AccessTokenSupplier produces access tokens using a GoogleCredential. It shouldn't need to know about an HttpRequestInitializer.

core/src/test/java/com/google/cloud/sql/core/DefaultAccessTokenSupplierTest.java
```diff
@@ -427,4 +440,48 @@ public void testGetTokenExpiration() {
...
+
+  private static class Oauth2BadCredentialFactory implements CredentialFactory {
+    private final OAuth2Credentials credential;
+
+    private Oauth2BadCredentialFactory(OAuth2Credentials credential) {
+      this.credential = credential;
+    }
+
+    @Override
+    public HttpRequestInitializer create() {
```
hessjcg, line 465: This simulates an old implementation of CredentialFactory from before `getCredentials()` was introduced that returns the wrong kind of credential in an HttpRequestInitializer.

core/src/test/java/com/google/cloud/sql/core/DefaultAccessTokenSupplierTest.java
```diff
@@ -427,4 +440,48 @@ public void testGetTokenExpiration() {
...
+      return new HttpCredentialsAdapter(credential);
+    }
+  }
+
+  private static class GoogleCredentialsFactory implements CredentialFactory {
```
hessjcg, line 470: This simulates the up-to-date implementation of CredentialFactory found in ApplicationDefaultCredentialFactory.

docs/jdbc-mariadb.md
```diff
@@ -96,6 +96,52 @@ Note: a non-empty string value for the `password` property must be set. While th
...
+account is a delegate to the previous service account. When delegation is
+used, each delegate must have the permissions named above on the service
+account it is delegating to.
+
+For example:
+```java
+    connProps.setProperty("delegates", "SERVICE_ACCOUNT_1,SERVICE_ACCOUNT_2,SERVICE_ACCOUNT_3");
+```
+
+In this example, the environment's IAM principal impersonates
```
hessjcg, line 141: I copied and adapted the cloud-sql-proxy help docs.

r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -38,6 +41,7 @@ public abstract class GcpConnectionFactoryProvider implements ConnectionFactoryP
   public static final Option<String> UNIX_SOCKET = Option.valueOf("UNIX_SOCKET");
   public static final Option<String> IP_TYPES = Option.valueOf("IP_TYPES");
   public static final Option<Boolean> ENABLE_IAM_AUTH = Option.valueOf("ENABLE_IAM_AUTH");
+  public static final Option<String> DELEGATES = Option.valueOf("DELEGATES");
```
hessjcg, line 44: Introduces a new R2DBC driver option for delegates.

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
enocom, line 42: Can we put some unit tests in place for this class? I'd like to see that the interface works before this method was added, and a test with it in use.

enocom (commented):
e2e impersonation tests will be critical here

core/src/main/java/com/google/cloud/sql/CredentialFactory.java
hessjcg: Sure. Existing tests are in DefaultAccessTokenSupplierTest. I'll move them into a test class specifically for this interface.

core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override
+  public GoogleCredentials getCredentials() {
+    GoogleCredentials credentials = source.getCredentials();
+
+    credentials =
+        ImpersonatedCredentials.newBuilder()
+            .setSourceCredentials(credentials)
+            .setTargetPrincipal(targetPrincipal)
+            .setDelegates(this.delegates)
```
enocom, line 92: Should we conditionally set delegates? What happens if delegates is a list of size 1?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -287,7 +299,7 @@ private static String getUserAgents() {
   /** Returns the current User-Agent header set for the underlying SQLAdmin API client. */
   private static String getApplicationName() {
     if (coreSocketFactory != null) {
-      return coreSocketFactory.adminApiService.getApplicationName();
+      return getUserAgents();
```
enocom, line 302: Is this change related to service account impersonation?

enocom (commented):
This PR has gotten too big.

Can we send the CredentialFactory interface change as one PR? Are there other things we can break out of this so we can make it smaller?

hessjcg:
Closing, replacing with 6 smaller PRs.

Issue #1168, opened 2023-02-20
Add service account impersonation


fzakaria:
### Feature Description

In order to support P4SA setup for Google Cloud, it would be great to add the ability to easily do service account impersonation in a common way across all connectors.

Briefly looking at the MySQL connector, I think this can be accomplished by writing a new custom GoogleCredentialFactory ourselves. Given the popularity of likely wanting to provide service account impersonation, it would be great to include it default in the standard factory.

### Alternatives Considered

We are considering writing our own factory for now.

enocom:
We're working on a v2 of the Java Connector and this is one of the features we want to include.

enocom:
On some further discussion, we're going to work on this and get some refactoring done as part of it and land this as part of v1 if it doesn't necessitate any breaking changes.

PR #1425, opened 2023-08-15
feat: Add service account impersonation credentials factory.

hessjcg:
Creates a new kind of CredentialFactory that can impersonate credentials. 

Part of #1168

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.services.sqladmin.SQLAdminScopes;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.ImpersonatedCredentials;
+import com.google.cloud.sql.CredentialFactory;
+import com.google.common.collect.Lists;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Wraps an existing CredentialFactory, adding service account impersonation for the list of
+ * delegates.
+ *
+ * <p>The "delegates" property is built to work like the cloud-sql-proxy
+ * --impersonate-service-account and gcloud --impersonate-service-account flags: The first element
+ * in the list is the target service account. Intermediate delegated credentials are applied first
+ * from end of the list, working towards the beginning of the list. For example, if <code>
+ * delegates = Arrays.asList( "first@serviceaccount.com", "second@serviceaccount.com",
+ * "third@serviceaccount.com")
+ * </code>
+ *
+ * <p>The connector will start with the GoogleCredentials supplied by `source`. It will use those to
+ * impersonate "third@serviceaccount.com", then with a token for "third@serviceaccount.com",
+ * impersonate "second@serviceaccount.com", and then impersonate "first@serviceaccount.com".
+ * Finally, the connector will attempt to access the services as "first@serviceaccount.com".
+ */
+class ServiceAccountImpersonatingCredentialFactory implements CredentialFactory {
+
+  private final CredentialFactory source;
+  private final List<String> delegates;
+  private final String targetPrincipal;
+
+  ServiceAccountImpersonatingCredentialFactory(CredentialFactory source, List<String> delegates) {
+    if (delegates == null || delegates.isEmpty()) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
```diff
@@ -0,0 +1,98 @@
...
+    //    targetPrincipal – the service account to impersonate
+    //    delegates – the chained list of delegates required to grant the final access_token.
+    //      If set, the sequence of identities must have "Service Account Token Creator"
+    //      capability granted to the preceding identity. For example, if set to
+    //      [serviceAccountB, serviceAccountC], the sourceCredential must have the Token Creator
+    //      role on serviceAccountB. serviceAccountB must have the Token Creator on
+    //      serviceAccountC. Finally, C must have Token Creator on target_principal.
+    //      If unset, sourceCredential must have that role on targetPrincipal.
+    List<String> reversedDelegates = Lists.reverse(delegates);
+    this.targetPrincipal = reversedDelegates.get(reversedDelegates.size() - 1);
```
enocom, line 74: Should this just be `reversedDelegates.get(0)`?

core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
```diff
@@ -0,0 +1,98 @@
...
+  public GoogleCredentials getCredentials() {
+    GoogleCredentials credentials = source.getCredentials();
+
+    credentials =
+        ImpersonatedCredentials.newBuilder()
+            .setSourceCredentials(credentials)
+            .setTargetPrincipal(targetPrincipal)
+            .setDelegates(this.delegates)
+            .setScopes(
+                Arrays.asList(SQLAdminScopes.SQLSERVICE_ADMIN, SQLAdminScopes.CLOUD_PLATFORM))
```
enocom, line 94: How does this work with Auto IAM Login where the sqlservice.login scope is required?

core/src/test/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    // Test that the CredentialsFactory.create() works.
+    Credentials impersonatedCredentials = newCredentials(impersonatedFactory);
+    assertThat(impersonatedCredentials).isInstanceOf(ImpersonatedCredentials.class);
+
+    // Test that CredentialsFatory.getCredentials() works.
+    assertThat(impersonatedFactory.getCredentials()).isInstanceOf(ImpersonatedCredentials.class);
+    ImpersonatedCredentials ic = (ImpersonatedCredentials) impersonatedFactory.getCredentials();
+    assertThat(ic.getAccount()).isEqualTo("first@serviceaccount.com");
+    assertThat(ic.getSourceCredentials()).isEqualTo(credentials);
```
enocom, line 54: Can we add an assertion here that checks the delegates?

Also shall we add a test for when delegates is empty?

core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
hessjcg: No. 
- First, this method receives `delegates` in accordance with the `--impersonate-service-account` contract, so  that the target credential is the first element.
- Next we reverse the list so that it is in order according to the Java ImpersonatedCredentials contract. The target credential is now the last element in `reversedDelegates`. 
- Then set this`this.targetPrincipal` to the last element in `reversedDelegates`
- Finally we set `this.delegates` to the beginning of the list, excluding the last element.

core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override
+  public GoogleCredentials getCredentials() {
+    GoogleCredentials credentials = source.getCredentials();
+
+    credentials =
+        ImpersonatedCredentials.newBuilder()
+            .setSourceCredentials(credentials)
+            .setTargetPrincipal(targetPrincipal)
+            .setDelegates(this.delegates)
```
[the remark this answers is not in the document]
hessjcg: Yes

core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
hessjcg: To create a valid ImpersonatedCredential, you must set the scopes, but we don't need to set the `sqladmin.login` scope right now.

If IAM is enabled, the DefaultAccessTokenSupplier will use this ImpersonatedCredential instance to create downscoped credentials with a `sqladmin.login` scope.

core/src/test/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactoryTest.java
hessjcg: Unfortunately, we can't test delegates. ImpersonatedCredentials does not expose delegates in its public api.

I'll add a test case where `delegates` contains a single element.

core/src/main/java/com/google/cloud/sql/core/ServiceAccountImpersonatingCredentialFactory.java
enocom: Sorry I misspoke. I mean, should this just be `delegates.get(0)`?

hessjcg: Well, that would return the same thing. But I think it is more straightforward to only use reversedDelegates.

enocom (approved):

PR #1433, opened 2023-08-17
fix: do not discard current instance info on force refresh.

akshatriya-01:
This commit changes the force-refresh behavior in that the current instance details will be retained in case the connection issues are caused by temporary network or instance availability issues. When the refresh operation completes, it will update the connection info with a refreshed value.


enocom:
We'll take this PR over. Thanks for getting us started.

PR #1441, opened 2023-08-24
fix: re-use existing connection info during force refresh

### LINKED ISSUE #1450: Re-use connection info during ZDT
### Feature Description

When a Cloud SQL instance experiences ZDT (zonal down time), there is a temporary period (< 5s) where the instance isn't available. For any connection that comes in during that period, the Java Connector will presently fail to connect and then trigger a refresh cycle. When this happens multiple times, the refresh cycle will be rate limited (1 refresh / 30s) which means the connection will not work for another 30s. To make matters worse, this refresh cycle will discard otherwise valid connection info.

To improve this situation, the Java Connector should not discard the current connection info and instead schedule the refresh in the background, such that in the case of ZDT MTTR is much faster since current will work again after a few seconds.

### Sample code

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

enocom:
When the connector initiates a force refresh, it would previously discard existing connection info and block on the refresh operation completing. In case the cause of the connection failure is transient, we keep the current connection info around and refresh in the background. When the current connection info is in fact invalid, this will cause callers to see additional failures until the refresh completes (usually < 1s).

Fixes #1450.


core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -138,13 +138,16 @@ public void testCloudSqlInstanceForcesRefresh() throws Exception {
...
+    assertThat(refreshCount.get()).isEqualTo(1);
 
-    SslData gotSslData = instance.getSslData();
-    assertThat(gotSslData).isSameInstanceAs(sslData);
     instance.forceRefresh();
+
     instance.getSslData();
-    assertThat(refreshCount.get()).isEqualTo(2);
+    // refresh count hasn't changed because we re-use the existing connection info
+    assertThat(refreshCount.get()).isEqualTo(1);
```
hessjcg, line 150: We need to add on to this test that forceRefresh() still works. That eventually `assertThat(refreshCount.get()).isEqualTo(2);` is true.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceConcurrencyTest.java
```diff
@@ -52,91 +51,7 @@ public void initialize(HttpRequest var1) throws IOException {
     }
   }
 
-  @Test(timeout = 20000)
-  public void testThatForceRefreshBalksWhenARefreshIsInProgress() throws Exception {
```
enocom, line 56: I think this test was built around the old refresh behavior -- where current would block on the next refresh. I need to double check that we have good coverage for balking before I ask for another review.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -111,41 +112,50 @@ public void testInstanceFailsOnConnectionError() throws Exception {
...
+
+    for (int i = 0; i < 10; i++) {
+      instance.getSslData();
+      if (refreshCount.get() > 1) {
+        return;
+      }
+      Thread.sleep(100);
+    }
+
+    fail(String.format("refresh count should be 2, got = %d", refreshCount.get()));
```
hessjcg, line 158: This works now.

hessjcg (approved):

PR #1457, opened 2023-09-08
fix: Improve handling of futures and threads during refresh.

hessjcg:
Rewrite the performRefresh() as a chain of task futures from the ListeningScheduledExecutorService. Now, tasks
submitted to the ListeningScheduledExecutorService never block on another task submitted to the ListeningScheduledExecutorService.

This should fix a category of bugs that show up in exceptions and logs as "connection timed out" or "refresh failed"
or "bad client certificate". These exceptions can occur when the credentials fail to refresh. 

This is the underlying bug: The ListeningScheduledExecutorService gets into a state where all its threads are busy 
running tasks, all running tasks are blocked waiting for recently submitted task to complete, and the recently
submitted tasks can't start because there are no available threads in the ListeningScheduledExecutorService.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.security.KeyPair;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -46,7 +47,8 @@ import javax.net.ssl.SSLSocket;
 class CloudSqlInstance {
 
   private static final Logger logger = Logger.getLogger(CloudSqlInstance.class.getName());
-  public static final int MAX_INSTANCE_DATA_WAIT_SEC = 30;
+
+  public static final int MAX_INSTANCE_DATA_WAIT_MS = 30000;
 
   private final ListeningScheduledExecutorService executor;
   private final InstanceDataSupplier instanceDataSupplier;
@@ -110,26 +112,59 @@ class CloudSqlInstance {
 
   /**
    * Returns the current data related to the instance from {@link #startRefreshAttempt()}. May block
-   * if no valid data is currently available.
+   * if no valid data is currently available. This method is called by an application thread when it
+   * is trying to create a new connection to the database. (It is not called by a
+   * ListeningScheduledExecutorService task.) So it is OK to block waiting for a future to complete.
+   *
+   * <p>When no refresh attempt is in progress, this returns immediately. Otherwise, it waits up to
+   * 30 seconds. If a refresh attempt succeeds, returns immediately at the end of that successful
+   * attempt. If no attempts succeed within the 30 second timeout, throws a RuntimeException with
+   * the exception from the last failed refresh attempt as the cause.
+   *
+   * <p>The behavior of this method has changed from earlier releases. In version 1.14.1 and
+   * earlier, When no refresh attempt is in progress, returns immediately. Otherwise, blocks
+   * application thread until the current refresh attempt finishes. If the refresh attempt succeeds,
+   * this returns the InstanceData. If not, this throws a RuntimeException, while a new refresh
+   * attempt is submitted to the executor in the background.
    */
-  private InstanceData getInstanceData() {
+  private InstanceData getInstanceData(long timeoutMs) {
     ListenableFuture<InstanceData> instanceDataFuture;
     synchronized (instanceDataGuard) {
       instanceDataFuture = currentInstanceData;
     }
     try {
-      return instanceDataFuture.get(MAX_INSTANCE_DATA_WAIT_SEC, TimeUnit.SECONDS);
+
+      // If the currentInstanceData has expired, then force refresh (which will balk if a refresh
+      // is already running) and make this and future requests to getInstanceData wait on the
+      // refresh operation to complete.
+      if (instanceDataFuture.isDone()) {
+        Date expiration = instanceDataFuture.get().getExpiration();
+        if (expiration.before(new Date())) {
+          forceRefresh();
+          synchronized (instanceDataGuard) {
+            currentInstanceData = nextInstanceData;
+            instanceDataFuture = currentInstanceData;
+          }
+        }
+      }
+
+      return instanceDataFuture.get(timeoutMs, TimeUnit.MILLISECONDS);
     } catch (TimeoutException e) {
       synchronized (instanceDataGuard) {
         if (currentRefreshFailure != null) {
           throw new RuntimeException(
-              "Unable to get valid instance data within 30 seconds. Last refresh attempt failed:"
+              String.format(
+                      "Unable to get valid instance data within %d ms. Last refresh attempt failed:",
+                      timeoutMs)
                   + currentRefreshFailure.getMessage(),
               currentRefreshFailure);
         }
       }
       throw new RuntimeException(
-          "Unable to get valid instance data within 30 seconds. No refresh has completed.", e);
+          String.format(
+              "Unable to get valid instance data within %d ms. No refresh has completed.",
+              timeoutMs),
+          e);
     } catch (ExecutionException | InterruptedException ex) {
       Throwable cause = ex.getCause();
       Throwables.throwIfUnchecked(cause);
@@ -142,7 +177,11 @@ class CloudSqlInstance {
    * block until required instance data is available.
    */
   SSLSocket createSslSocket() throws IOException {
-    return (SSLSocket) getInstanceData().getSslContext().getSocketFactory().createSocket();
+    return (SSLSocket)
+        getInstanceData(MAX_INSTANCE_DATA_WAIT_MS)
+            .getSslContext()
+            .getSocketFactory()
+            .createSocket();
   }
 
   /**
@@ -156,7 +195,7 @@ class CloudSqlInstance {
    *     preferences.
    */
   String getPreferredIp(List<String> preferredTypes) {
-    Map<String, String> ipAddrs = getInstanceData().getIpAddrs();
+    Map<String, String> ipAddrs = getInstanceData(MAX_INSTANCE_DATA_WAIT_MS).getIpAddrs();
     for (String ipType : preferredTypes) {
       String preferredIp = ipAddrs.get(ipType);
       if (preferredIp != null) {
@@ -198,6 +237,10 @@ class CloudSqlInstance {
    * future that resolves once a valid InstanceData has been acquired. This sets up a chain of
    * futures that will 1. Acquire a rate limiter. 2. Attempt to fetch instance data. 3. Schedule the
    * next attempt to get instance data based on the success/failure of this attempt.
+   *
+   * <p>This behavior changed. In version 1.14.1 and earlier, this method was called
+   * performRefresh(). It used to block until either (1) an InstanceData was retrieved or (2) an
+   * attempt failed, and an exception was thrown.
    *
    * @see com.google.cloud.sql.core.CloudSqlInstance#handleRefreshResult(
    *     com.google.common.util.concurrent.ListenableFuture)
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -113,8 +116,19 @@ private InstanceData getInstanceData() {
       instanceDataFuture = currentInstanceData;
     }
     try {
-      return Uninterruptibles.getUninterruptibly(instanceDataFuture);
-    } catch (ExecutionException ex) {
+      return instanceDataFuture.get(30, TimeUnit.SECONDS);
```
hessjcg, line 119: This is a small change of behavior: 

The old way hangs application thread until the current request finishes. If the request succeeds, this returns the InstanceData. If not, this throws a RuntimeException. 

The new way waits up to 30 seconds until a refresh attempt succeeds. If a refresh attempt succeeds within the 30 second timeout, the a future containing the latest InstanceData is returned. If not, this throws a RuntimeException using the exception from the last failed refresh attempt.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -173,30 +187,63 @@ void forceRefresh() {
...
+   * futures that will 1. Acquire a rate limiter. 2. Attempt to fetch instance data. 3. Schedule the
+   * next attempt to get instance data based on the success/failure of this attempt.
+   *
+   * @see com.google.cloud.sql.core.CloudSqlInstance#rescheduleAfterGetInstanceData(
+   *     com.google.common.util.concurrent.ListenableFuture)
    */
-  private InstanceData performRefresh() throws InterruptedException, ExecutionException {
-    logger.fine(
-        String.format("[%s] Refresh Operation: Acquiring rate limiter permit.", instanceName));
+  private ListenableFuture<InstanceData> performRefresh() {
```
hessjcg, line 204: This is a change of behavior. 

The old way: performRefresh() used to block until either (1) an InstanceData was retrieved or (2) an attempt failed, and an exception was thrown. 

The new way: performRefresh() does not block. It sets up a chain of futures. The future returned from this method will resolve only when a refresh attempt succeeds.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -111,12 +111,49 @@ public void testInstanceFailsOnConnectionError() throws Exception {
...
             keyPairFuture,
-            RateLimiter.create(1.0 / 30.0));
+            RateLimiter.create(10));
 
     RuntimeException ex = Assert.assertThrows(RuntimeException.class, instance::getSslData);
     assertThat(ex).hasMessageThat().contains("always fails");
   }
 
+  @Test
+  public void testInstanceFailsOnTooLongToRetrieve() throws Exception {
```
hessjcg, line 121: A new test covers the behavior when the refresh attempt never returns.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -139,7 +176,7 @@ public void testCloudSqlInstanceForcesRefresh() throws Exception {
             stubCredentialFactory,
             executorService,
             keyPairFuture,
-            RateLimiter.create(1.0 / 30.0));
+            RateLimiter.create(10));
```
hessjcg, line 179: Speed up tests - allow 10 requests/second.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
enocom: Minor not: could we pull out the 30 into a constant? `MAX_INSTANCE_DATA_WAIT_TIME` or similar?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -216,13 +264,22 @@ private InstanceData performRefresh() throws InterruptedException, ExecutionExce
...
-            executor.schedule(this::performRefresh, secondsToRefresh, TimeUnit.SECONDS);
         // Refresh completed successfully, reset forceRefreshRunning.
         forceRefreshRunning = false;
+        currentRefreshFailure = null;
+        currentInstanceData = Futures.immediateFuture(data);
+
+        // Now update nextInstanceData to perform a refresh after the
+        // scheduled delay
+        ListenableFuture scheduleDelay =
+            executor.schedule(() -> {}, secondsToRefresh, TimeUnit.SECONDS);
```
enocom, line 275: Is there any other ways to do this? On the surface it looks unusual.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -111,12 +111,49 @@ public void testInstanceFailsOnConnectionError() throws Exception {
...
+          ListenableFuture<?> f =
+              exec.submit(
+                  () -> {
+                    try {
+                      Thread.sleep(120 * 1000); // stop for 2 minutes
+                    } catch (InterruptedException e) {
+                    }
+                    throw new RuntimeException("fake read timeout");
+                  });
+          f.get(); // this will throw an ExecutionException
```
enocom, line 138: Let's add some more details to the comment -- it will throw an exception because we've configured the connector wait only 30s?

enocom (commented):
This looks like a nice improvement.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: Done.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: I found a better method call for this.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: I added comments and cleaned this up.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -119,18 +116,55 @@ public void testInstanceFailsOnConnectionError() throws Exception {
...
+    SslData sslData = new SslData(null, null, null);
+    InstanceData data =
+        new InstanceData(null, sslData, Date.from(Instant.now().plus(1, ChronoUnit.HOURS)));
     AtomicInteger refreshCount = new AtomicInteger();
 
     CloudSqlInstance instance =
         new CloudSqlInstance(
             "project:region:instance",
             (instanceName, accessTokenSupplier, authType, executor, keyPair) -> {
+              Thread.sleep(100);
```
enocom, line 165: Can we avoid the sleep somehow? These are notoriously flaky otherwise.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -99,23 +103,34 @@ class CloudSqlInstance {
...
+      synchronized (instanceDataGuard) {
+        if (currentRefreshFailure != null) {
+          throw new RuntimeException(
+              "Unable to get valid instance data within 30 seconds. Last refresh attempt failed:"
+                  + currentRefreshFailure.getMessage(),
+              currentRefreshFailure);
+        }
+      }
+      throw new RuntimeException(
+          "Unable to get valid instance data within 30 seconds. No refresh has completed.", e);
```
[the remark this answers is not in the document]
enocom: Same above.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: I've updated these tests (and added a few new ones) to use a lock instead of waiting.

enocom:
@hessjcg Let's get the tests green and then we can get this merged

enocom:
@hessjcg friendly ping on this. This is an important fix and improvement that we want to bake in CI to see if it resolves the flaky tests.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -47,6 +47,8 @@ class CloudSqlInstance {
 
   private static final Logger logger = Logger.getLogger(CloudSqlInstance.class.getName());
 
+  public static final int MAX_INSTANCE_DATA_WAIT_MS = 30000;
```
enocom, line 50: Let's add a comment explaining what this constant represents.

pom.xml
```diff
@@ -153,6 +153,7 @@
           <compilerArgs>
             <arg>-XDcompilePolicy=simple</arg>
             <arg>-Xplugin:ErrorProne</arg>
+            <arg>-Xlint:-options</arg>
```
enocom, line 156: Should this be a separate PR?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -159,45 +216,88 @@ String getPreferredIp(List<String> preferredTypes) {
    * new refresh is already in progress. If successful, other methods will block until refresh has
    * been completed.
    */
-  void forceRefresh() {
+  boolean forceRefresh() {
```
enocom, line 219: Should this be returning a boolean? I don't see it being used.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -99,23 +104,71 @@ class CloudSqlInstance {
...
     synchronized (instanceDataGuard) {
       instanceDataFuture = currentInstanceData;
+      // If the currentInstanceData has expired, then force refresh (which will balk if a refresh
+      // is already running) and make this and future requests to getInstanceData wait on the
+      // refresh operation to complete.
+      if (instanceDataFuture.isDone()) {
+        Instant expiration;
+        try {
+          expiration = instanceDataFuture.get().getExpiration();
+          if (expiration == null || expiration.isBefore(Instant.now())) {
```
enocom, line 140: If the current instance data has already expired, won't we ended up initiating a force refresh when the socket creation fails? In other words, do we really need to do this work here?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -99,23 +104,71 @@ class CloudSqlInstance {
...
+      }
     }
+
     try {
-      return Uninterruptibles.getUninterruptibly(instanceDataFuture);
-    } catch (ExecutionException ex) {
+      return instanceDataFuture.get(timeoutMs, TimeUnit.MILLISECONDS);
+    } catch (TimeoutException e) {
+      synchronized (instanceDataGuard) {
+        if (currentRefreshFailure != null) {
```
enocom, line 156: Could we split this out as a separate PR? It's nice to surface the refresh failure like this.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -159,45 +216,88 @@ String getPreferredIp(List<String> preferredTypes) {
...
    */
-  void forceRefresh() {
+  boolean forceRefresh() {
     synchronized (instanceDataGuard) {
-      // Don't force a refresh until the current forceRefresh operation
+      // Don't force a refresh until the current refresh operation
       // has produced a successful refresh.
-      if (forceRefreshRunning) {
-        return;
+      if (refreshRunning) {
```
enocom, line 223: Could this rename be a separate PR? Seems like a nice improvement.

hessjcg:
Closing, too big. I'll submit this as a sequence of PRs.

PR #1544, opened 2023-09-21
chore: fix lint warnings

### LINKED ISSUE #1460: Correct all lint warnings in build
Also, we should consider failing the build instead of just warning on lint problems.

ttosta-google:
Fixes #1460

Notes:

* Now the build fails on warnings.
* `Thread.yield();` triggers the warning [ThreadPriorityCheck](https://errorprone.info/bugpattern/ThreadPriorityCheck). So I replaced it by `Thread.sleep(0);` as they may call same system calls in some platforms. Please let me know if you have any other suggestions.

### DIFF
--- a/jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
+++ b/jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
@@ -18,12 +18,14 @@ package com.google.cloud.sql.sqlserver;
 
 import com.google.cloud.sql.core.CoreSocketFactory;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Splitter;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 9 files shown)


jdbc/sqlserver/src/main/java/com/google/cloud/sql/sqlserver/SocketFactory.java
```diff
@@ -46,21 +48,23 @@ public class SocketFactory extends javax.net.SocketFactory {
    * connections to a Cloud SQL instance.
    */
   public SocketFactory(String socketFactoryConstructorArg) throws UnsupportedEncodingException {
-    String[] s = socketFactoryConstructorArg.split("\\?");
-    this.props.setProperty(CoreSocketFactory.CLOUD_SQL_INSTANCE_PROPERTY, s[0]);
-    if (s.length == 2 && s[1].length() > 0) {
-      String[] queryParams = s[1].split("&");
+    List<String> s = Splitter.on('?').splitToList(socketFactoryConstructorArg);
```
enocom, line 51: Do we have any unit tests covering this code?

We should ensure we have thorough testing of this code before changing it here, just out of caution. Also, once we have those tests in place, we should clean this code up -- on the surface it looks more complicated than it needs to be.

ttosta-google: Yes, we have unit tests for this code in [JdbcSqlServerUnitTests.java](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/jdbc/sqlserver/src/test/java/com/google/cloud/sql/sqlserver/JdbcSqlServerUnitTests.java).

I will create a new issue to refactor this code.

enocom (approved):
LGTM

Could we port the fail on lint warnings to AlloyDB Java too?

PR #1571, opened 2023-10-03
test: Testcase improvements for the CloudSqlInstanceTest

hessjcg:
Improves the stability of the CloudSqlInstanceTest and CloudSqlInstanceConcurrencyTest. This adds testcases
to cover refresh behavior for tokens near expiration. Also, it adds the PauseCondition so that the testcase can
orchestrate the refresh behavior across threads without resorting to `Thread.sleep()`.

### DIFF
--- a/core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
+++ b/core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
@@ -16,7 +16,6 @@
 package com.google.cloud.sql.core;
 
 import static com.google.common.truth.Truth.assertThat;
-import static org.junit.Assert.fail;
 
 import com.google.cloud.sql.AuthType;
 import com.google.common.collect.ImmutableMap;
@@ -89,13 +88,7 @@ public class CloudSqlInstanceTest {
             AuthType authType,
             ListeningScheduledExecutorService exec,
             ListenableFuture<KeyPair> keyPair) -> {
-          ListenableFuture<?> f =
-              exec.submit(
-                  () -> {
-                    throw new RuntimeException("always fails");
-                  });
-          f.get(); // this will throw an ExecutionException
-          return null;
+          throw new RuntimeException("always fails");
         };
 
     // initialize instance after mocks are set up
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -134,21 +135,150 @@ public void testCloudSqlInstanceForcesRefresh() throws InterruptedException {
...
+    // getSslData again, and assert the refresh operation completed.
+    SslData d2 = instance.getSslData();
+    assertThat(d2).isSameInstanceAs(data.getSslData());
+  }
+
+  @Test
+  public void testThatForceRefreshBalksWhenAForceRefreshIsInProgress() throws Exception {
+    InstanceData aboutToExpireData =
+        new InstanceData(
+            null, new SslData(null, null, null), Instant.now().plus(1, ChronoUnit.HOURS));
```
ttosta-google, line 219: it should be `ChronoUnit.SECONDS`

enocom (commented):
In AlloyDB Java we use a deterministic scheduler to write tests like this. Could we do the same here?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: In this testcase, we want the initial data to have a long expiration, so that it does not immediately trigger a refresh. Then, we want to call forceRefresh() and see that exactly 1 refresh operation occurs. 

I have renamed the variable `initialData` instead of `aboutToExpireData` for clarity.

hessjcg:
After some consideration, I think we should test concurrency using a ScheduledExecutor, while controlling the flow of execution using the PauseCondition. This approach actually exercises the concurrency behavior. Also, since our tests run across many OS and VM versions, it becamse easy to spot and fix subtle concurrency programming problems because a testcase would fail on only one or two OS+VM combo but not the others.

pom.xml
```diff
@@ -153,6 +153,7 @@
           <compilerArgs>
             <arg>-XDcompilePolicy=simple</arg>
             <arg>-Xplugin:ErrorProne</arg>
+            <arg>-Xlint:-options</arg> <!-- ignore warning about deprecated java 8 src -->
```
enocom, line 156: nit: this would be nicer as separate PR just so people can discover it more easily.

enocom (approved):

pom.xml
hessjcg: I'll move it.

enocom:
One other thought: we should port this to AlloyDB Java.

pom.xml
hessjcg: See #1582

PR #1573, opened 2023-10-03
fix: Improve handling of futures and threads during refresh.

hessjcg:
Rewrite the performRefresh() as a chain of task futures from the ListeningScheduledExecutorService. Now, tasks
submitted to the ListeningScheduledExecutorService never block on another task submitted to the ListeningScheduledExecutorService.

This should fix a category of bugs that show up in exceptions and logs as "connection timed out" or "refresh failed"
or "bad client certificate". These exceptions can occur when the credentials fail to refresh.

This is the underlying bug: The ListeningScheduledExecutorService gets into a state where all its threads are busy
running tasks, all running tasks are blocked waiting for recently submitted task to complete, and the recently
submitted tasks can't start because there are no available threads in the ListeningScheduledExecutorService.

This changes the behavior of CloudSqlInstance.getInstanceData() and CloudSqlInstance.startRefreshAttempt()
in ways that have a very small possibility of destabilizing customer applications. 

In version 1.14.1 and earlier:  CloudSqlInstance.getInstanceData() behaved like this: When no refresh attempt is 
in progress, returns immediately. Otherwise, blocks application thread until the current refresh attempt finishes. 
If the refresh attempt succeeds, this returns the InstanceData. If not, this throws a RuntimeException, while a 
new refresh attempt is submitted to the executor in the background.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
@@ -23,7 +23,6 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningScheduledExecutorService;
 import com.google.common.util.concurrent.RateLimiter;
-import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import java.io.IOException;
 import java.security.KeyPair;
@@ -68,7 +67,10 @@ class CloudSqlInstance {
   private ListenableFuture<InstanceData> nextInstanceData;
 
   @GuardedBy("instanceDataGuard")
-  private boolean forceRefreshRunning;
+  private boolean refreshRunning;
+
+  @GuardedBy("instanceDataGuard")
+  private Throwable currentRefreshFailure;
 
   /**
    * Initializes a new Cloud SQL instance based on the given connection name.
@@ -100,26 +102,48 @@ class CloudSqlInstance {
     }
 
     synchronized (instanceDataGuard) {
-      this.currentInstanceData = executor.submit(this::performRefresh);
+      this.currentInstanceData = this.startRefreshAttempt();
       this.nextInstanceData = currentInstanceData;
     }
   }
 
   /**
-   * Returns the current data related to the instance from {@link #performRefresh()}. May block if
-   * no valid data is currently available.
+   * Returns the current data related to the instance from {@link #startRefreshAttempt()}. May block
+   * if no valid data is currently available. This method is called by an application thread when it
+   * is trying to create a new connection to the database. (It is not called by a
+   * ListeningScheduledExecutorService task.) So it is OK to block waiting for a future to complete.
+   *
+   * <p>When no refresh attempt is in progress, this returns immediately. Otherwise, it waits up to
+   * timeoutMs milliseconds. If a refresh attempt succeeds, returns immediately at the end of that
+   * successful attempt. If no attempts succeed within the timeout, throws a RuntimeException with
+   * the exception from the last failed refresh attempt as the cause.
    */
   private InstanceData getInstanceData(long timeoutMs) {
     ListenableFuture<InstanceData> instanceDataFuture;
     synchronized (instanceDataGuard) {
       instanceDataFuture = currentInstanceData;
     }
+
     try {
-      return Uninterruptibles.getUninterruptibly(
-          instanceDataFuture, timeoutMs, TimeUnit.MILLISECONDS);
-    } catch (TimeoutException ex) {
-      throw new RuntimeException(ex);
-    } catch (ExecutionException ex) {
+      return instanceDataFuture.get(timeoutMs, TimeUnit.MILLISECONDS);
+    } catch (TimeoutException e) {
+      synchronized (instanceDataGuard) {
+        if (currentRefreshFailure != null) {
+          throw new RuntimeException(
+              String.format(
+                      "Unable to get valid instance data within %d ms."
+                          + " Last refresh attempt failed:",
+                      timeoutMs)
+                  + currentRefreshFailure.getMessage(),
+              currentRefreshFailure);
+        }
+      }
+      throw new RuntimeException(
+          String.format(
+              "Unable to get valid instance data within %d ms. No refresh has completed.",
+              timeoutMs),
+          e);
+    } catch (ExecutionException | InterruptedException ex) {
       Throwable cause = ex.getCause();
       Throwables.throwIfUnchecked(cause);
       throw new RuntimeException(cause);
@@ -165,43 +189,77 @@ class CloudSqlInstance {
    */
   void forceRefresh() {
     synchronized (instanceDataGuard) {
-      // Don't force a refresh until the current forceRefresh operation
+      // Don't force a refresh until the current refresh operation
       // has produced a successful refresh.
-      if (forceRefreshRunning) {
+      if (refreshRunning) {
         return;
       }
-
-      forceRefreshRunning = true;
       nextInstanceData.cancel(false);
       logger.fine(
           String.format(
               "[%s] Force Refresh: the next refresh operation was cancelled."
                   + " Scheduling new refresh operation immediately.",
               instanceName));
-      nextInstanceData = executor.submit(this::performRefresh);
+      nextInstanceData = this.startRefreshAttempt();
     }
   }
 
   /**
-   * Triggers an update of internal information obtained from the Cloud SQL Admin API. Replaces the
-   * value of currentInstanceData and schedules the next refresh shortly before the information
-   * would expire.
+   * Triggers an update of internal information obtained from the Cloud SQL Admin API, returning a
+   * future that resolves once a valid InstanceData has been acquired. This sets up a chain of
+   * futures that will 1. Acquire a rate limiter. 2. Attempt to fetch instance data. 3. Schedule the
+   * next attempt to get instance data based on the success/failure of this attempt.
+   *
+   * @see com.google.cloud.sql.core.CloudSqlInstance#handleRefreshResult(
+   *     com.google.common.util.concurrent.ListenableFuture)
    */
-  private InstanceData performRefresh() throws InterruptedException, ExecutionException {
-    logger.fine(
-        String.format("[%s] Refresh Operation: Acquiring rate limiter permit.", instanceName));
-    // To avoid unreasonable SQL Admin API usage, use a rate limit to throttle our usage.
-    //noinspection UnstableApiUsage
-    forcedRenewRateLimiter.acquire();
-    logger.fine(
-        String.format(
-            "[%s] Refresh Operation: Acquired rate limiter permit. Starting refresh...",
-            instanceName));
+  private ListenableFuture<InstanceData> startRefreshAttempt() {
+    // As soon as we begin submitting refresh attempts to the executor, mark a refresh
+    // as "in-progress" so that subsequent forceRefresh() calls balk until this one completes.
+    synchronized (instanceDataGuard) {
+      refreshRunning = true;
+    }
 
+    // To avoid unreasonable SQL Admin API usage, use a rate limit to throttle our usage.
+    ListenableFuture<?> rateLimit =
+        executor.submit(
+            () -> {
+              logger.fine(
+                  String.format(
+                      "[%s] Refresh Operation: Acquiring rate limiter permit.", instanceName));
+              //noinspection UnstableApiUsage
+              forcedRenewRateLimiter.acquire();
+              logger.fine(
+                  String.format(
+                      "[%s] Refresh Operation: Acquired rate limiter permit. Starting refresh...",
+                      instanceName));
+            },
+            executor);
+
+    // Once rate limiter is done, attempt to getInstanceData.
+    ListenableFuture<InstanceData> dataFuture =
+        Futures.whenAllComplete(rateLimit)
+            .callAsync(
+                () ->
+                    instanceDataSupplier.getInstanceData(
+                        this.instanceName,
+                        this.accessTokenSupplier,
+                        this.authType,
+                        executor,
+                        keyPair),
+                executor);
+
+    // Finally, reschedule refresh after getInstanceData is complete.
+    return Futures.whenAllComplete(dataFuture)
+        .callAsync(() -> handleRefreshResult(dataFuture), executor);
+  }
+
+  private ListenableFuture<InstanceData> handleRefreshResult(
+      ListenableFuture<InstanceData> dataFuture) {
     try {
-      InstanceData data =
-          instanceDataSupplier.getInstanceData(
-              this.instanceName, this.accessTokenSupplier, this.authType, executor, keyPair);
+      // This does not block, because it only gets called when dataFuture has completed.
+      // This will throw an exception if the refresh attempt has failed.
+      InstanceData data = dataFuture.get();
 
       logger.fine(
           String.format(
@@ -220,13 +278,21 @@ class CloudSqlInstance {
                   .toString()));
 
       synchronized (instanceDataGuard) {
-        currentInstanceData = Futures.immediateFuture(data);
-        nextInstanceData =
-            executor.schedule(this::performRefresh, secondsToRefresh, TimeUnit.SECONDS);
         // Refresh completed successfully, reset forceRefreshRunning.
-        forceRefreshRunning = false;
+        refreshRunning = false;
+        currentRefreshFailure = null;
+        currentInstanceData = Futures.immediateFuture(data);
+
+        // Now update nextInstanceData to perform a refresh after the
+        // scheduled delay
+        nextInstanceData =
+            Futures.scheduleAsync(
+                this::startRefreshAttempt, secondsToRefresh, TimeUnit.SECONDS, executor);
+
+        // Resolves to an InstanceData immediately
+        return currentInstanceData;
       }
-      return data;
+
     } catch (ExecutionException | InterruptedException e) {
       logger.log(
           Level.FINE,
@@ -235,9 +301,12 @@ class CloudSqlInstance {
               instanceName),
           e);
       synchronized (instanceDataGuard) {
-        nextInstanceData = executor.submit(this::performRefresh);
+        currentRefreshFailure = e;
+        nextInstanceData = this.startRefreshAttempt();
+
+        // Resolves after the next successful refresh attempt.
+        return nextInstanceData;
       }
-      throw e;
     }
   }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -100,26 +102,71 @@ class CloudSqlInstance {
...
+   * if no valid data is currently available. This method is called by an application thread when it
+   * is trying to create a new connection to the database. (It is not called by a
+   * ListeningScheduledExecutorService task.) So it is OK to block waiting for a future to complete.
+   *
+   * <p>When no refresh attempt is in progress, this returns immediately. Otherwise, it waits up to
+   * 30 seconds. If a refresh attempt succeeds, returns immediately at the end of that successful
+   * attempt. If no attempts succeed within the 30 second timeout, throws a RuntimeException with
+   * the exception from the last failed refresh attempt as the cause.
+   *
+   * <p>The behavior of this method has changed from earlier releases. In version 1.14.1 and
```
enocom, line 121: Do we want to add this? Why would this be useful to know in the long term?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -163,45 +210,88 @@ String getPreferredIp(List<String> preferredTypes, long timeoutMs) {
    * new refresh is already in progress. If successful, other methods will block until refresh has
    * been completed.
    */
-  void forceRefresh() {
+  boolean forceRefresh() {
```
enocom, line 213: Are we using this return value?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -100,26 +102,71 @@ class CloudSqlInstance {
...
+   * attempt is submitted to the executor in the background.
    */
   private InstanceData getInstanceData(long timeoutMs) {
     ListenableFuture<InstanceData> instanceDataFuture;
     synchronized (instanceDataGuard) {
       instanceDataFuture = currentInstanceData;
+      // If the currentInstanceData has expired, then force refresh (which will balk if a refresh
+      // is already running) and make this and future requests to getInstanceData wait on the
+      // refresh operation to complete.
+      if (instanceDataFuture.isDone()) {
```
enocom, line 134: I think we don't need this block of code (down to line 147).

Here's my reasoning.

If an instance data has an expired cert, the next connection attempt will fail and trigger a force refresh for us. So in effect, we have this functionality elsewhere in the Java Connector.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -163,45 +210,88 @@ String getPreferredIp(List<String> preferredTypes, long timeoutMs) {
...
   /**
-   * Triggers an update of internal information obtained from the Cloud SQL Admin API. Replaces the
-   * value of currentInstanceData and schedules the next refresh shortly before the information
-   * would expire.
+   * Triggers an update of internal information obtained from the Cloud SQL Admin API, returning a
+   * future that resolves once a valid InstanceData has been acquired. This sets up a chain of
+   * futures that will 1. Acquire a rate limiter. 2. Attempt to fetch instance data. 3. Schedule the
+   * next attempt to get instance data based on the success/failure of this attempt.
+   *
+   * <p>This behavior changed. In version 1.14.1 and earlier, this method was called
```
enocom, line 237: Would this be better in the body of the commit message? I can't think why we'd want it here otherwise.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -163,45 +210,88 @@ String getPreferredIp(List<String> preferredTypes, long timeoutMs) {
...
+   *     com.google.common.util.concurrent.ListenableFuture)
    */
-  private InstanceData performRefresh() throws InterruptedException, ExecutionException {
-    logger.fine(
-        String.format("[%s] Refresh Operation: Acquiring rate limiter permit.", instanceName));
+  private ListenableFuture<InstanceData> startRefreshAttempt() {
+    // As soon as we begin submitting refresh attempts to the executor, mark a refresh
+    // as "in-progress" so that subsequent forceRefresh() calls balk until this one completes.
+    synchronized (instanceDataGuard) {
+      refreshRunning = true;
```
enocom, line 248: Should we first check if refreshRunning was already set to true before doing this?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -163,45 +210,88 @@ String getPreferredIp(List<String> preferredTypes, long timeoutMs) {
...
+
+    // Finally, reschedule refresh after getInstanceData is complete.
+    ListenableFuture<InstanceData> rescheduleFuture =
+        Futures.whenAllComplete(dataFuture)
+            .callAsync(() -> handleRefreshResult(dataFuture), executor);
+
+    return rescheduleFuture;
+  }
 
+  private ListenableFuture<InstanceData> handleRefreshResult(
```
enocom, line 289: Why does this return a value?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -186,6 +190,46 @@ public void testCloudSqlInstanceForcesRefresh() throws Exception {
     assertThat(refreshCount.get()).isEqualTo(2);
   }
 
+  @Test
+  public void testCloudSqlInstanceRetriesOnInitialFailures() throws Exception {
+    InstanceData data =
+        new InstanceData(
+            null, new SslData(null, null, null), Instant.now().plus(1, ChronoUnit.HOURS));
```
enocom, line 197: Should we introduce test helpers to create these objects? All these nulls make it hard to see what we're doing otherwise.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -186,6 +190,46 @@ public void testCloudSqlInstanceForcesRefresh() throws Exception {
...
+
+    CloudSqlInstance instance =
+        new CloudSqlInstance(
+            "project:region:instance",
+            (instanceName, accessTokenSupplier, authType, executor, keyPair) -> {
+              int c = refreshCount.get();
+              refreshCount.incrementAndGet();
+              switch (c) {
+                case 0:
+                  refresh0.pause();
```
enocom, line 210: A quick comment on why we're pausing on the first request would be helpful here.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -186,6 +190,46 @@ public void testCloudSqlInstanceForcesRefresh() throws Exception {
...
+                  throw new RuntimeException("bad request 0");
+                case 1:
+                  throw new RuntimeException("bad request 1");
```
enocom, line 213: Why do we need two runtime exceptions? Would one suffice for this test?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -249,6 +294,77 @@ public void testCloudSqlRefreshesExpiredData() throws Exception {
         () -> instance.getSslData(TEST_TIMEOUT_MS) == data.getSslData(), 1000L);
   }
 
+  @Test
+  public void testThatForceRefreshBalksWhenAScheduledRefreshIsInProgress() throws Exception {
+    // Set expiration 1 minute in the future, so that it will trigger a scheduled refresh
+    // and won't expire during this testcase.
+    InstanceData aboutToExpireData =
+        new InstanceData(
+            null, new SslData(null, null, null), Instant.now().plus(1, ChronoUnit.MINUTES));
```
enocom, line 303: What if we pulled this into a helper method that took the arguments of `plus`? From there we can make it more extensible for other tests.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -249,6 +294,77 @@ public void testCloudSqlRefreshesExpiredData() throws Exception {
         () -> instance.getSslData(TEST_TIMEOUT_MS) == data.getSslData(), 1000L);
   }
 
+  @Test
+  public void testThatForceRefreshBalksWhenAScheduledRefreshIsInProgress() throws Exception {
+    // Set expiration 1 minute in the future, so that it will trigger a scheduled refresh
+    // and won't expire during this testcase.
+    InstanceData aboutToExpireData =
```
enocom, line 301: nit: expiresInOneMinute?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
enocom, line 303: Since we control time here, shall we just make this 5 seconds or something shorter to make our tests fast?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -249,6 +294,77 @@ public void testCloudSqlRefreshesExpiredData() throws Exception {
...
+    assertThat(refreshCount.get()).isEqualTo(1);
+    SslData d = instance.getSslData(TEST_TIMEOUT_MS);
+    assertThat(d).isSameInstanceAs(aboutToExpireData.getSslData());
+
+    // Because the data is about to expire, scheduled refresh will begin immediately.
+    // Wait until refresh is in progress.
+    refresh1.waitForPauseToStart(1000);
+
+    // Then call forceRefresh(), which should balk because a refresh attempt is in progress.
+    assertThat(instance.forceRefresh()).isFalse();
```
enocom, line 351: Since we're only using the boolean here in test, I think we shouldn't assert this way. Instead two calls to force refresh plus a check on the number of times it was called should be adequate, right?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -309,6 +425,105 @@ public void testThatForceRefreshBalksWhenAForceRefreshIsInProgress() throws Exce
     assertThat(refreshCount.get()).isEqualTo(2);
   }
 
+  @Test
+  public void testRefreshRetriesOnAfterFailedAttempts() throws Exception {
```
enocom, line 429: What lines of code are we covering in this test that we don't have covered above?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: I'll move this to the commit comment.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: I agree that a failed connection attempt forces a refresh. However, I think we still need this here to improve the
stability of our customer applications and tests. 

This PR makes the contract of this method more strict. This method will block until it has a *valid* InstanceData or throw an exception if the timeout expires. 

The old implementation would allow this bad scenario: 
1. Application request a database connection.
2. getInstanceData() returns an expired certificate.
3. CoreSocketFactory.createSocket() triggers a force refresh and throws an exception back to the application.
4. Application has to handle a failed database connection caused by our connector.

The new implementation now works like this: 
1. Application request a database connection
2. getInstanceData() blocks until the certificate is refreshed
4. Application continues uninterrupted. 

On the balance, I think this bit of code is worth the improved stability.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: I fixed the tests and changed this back to return void.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: Moved to commit message.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: No. That behavior is handled by forceRefresh()

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: Because we need to create a chain of futures to retry the refresh when an attempt fails. Let's discuss.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Good idea. Done.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: I guess this isn't strictly needed. I've updated the test accordingly.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Good point. Fixed.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Fixed.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Done.

hessjcg: I opted for 1 minute because I wanted to trigger a refresh, but I wanted to make sure the initial data does not expire during the test. This test takes less than 1 second to run on my laptop, but can take up to 5 seconds on some of the slower github action executors.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Good point. Fixed.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: This tests that `InstanceData getInstanceData(long timeoutMs)` will block until it has a valid InstanceData, 
even when currentInstanceData expires while a refresh attempt is in progress and subsequent refresh attempts fail. 

This is important for testing the way `handleRefreshResult()` chains futures together on failed 
refresh attempts. It also tests that `getInstanceData()` will block when the currentInstanceData has expired and a refresh attempt is in progress.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
enocom: Got it -- I didn't understand that callAsync takes a Future.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
enocom: If we still think this is important, let's pull it out into a separate PR. It seems like a separate concern from the bigger threading fixes here.

enocom (changes requested):

hessjcg:
New PR #1600 makes sure that `CloudSqlInstance.getInstanceData()` never returns expired data, but instead blocks
until it can return valid InstanceData.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -310,7 +403,80 @@ public void testThatForceRefreshBalksWhenAForceRefreshIsInProgress() throws Exce
...
   @Test
-  public void testGetPreferredIpTypes() throws Exception {
+  public void testRefreshRetriesOnAfterFailedAttempts() throws Exception {
+    InstanceData aboutToExpireData = newInstanceDataExpiringIn(10, ChronoUnit.MILLIS);
+
+    InstanceData data = newFutureInstanceData();
+
+    AtomicInteger refreshCount = new AtomicInteger();
+    final PauseCondition badRequest1 = new PauseCondition();
+    final PauseCondition badRequest2 = new PauseCondition();
```
enocom, line 413: Why two bad requests? Could we write this test with just one?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
```diff
@@ -310,7 +403,80 @@ public void testThatForceRefreshBalksWhenAForceRefreshIsInProgress() throws Exce
...
+    // Allow the second refresh operation to complete
+    badRequest1.proceed();
+    badRequest1.waitForPauseToEnd(5000);
+    badRequest1.waitForCondition(() -> refreshCount.get() == 2, 2000);
+
+    // Allow the second bad request completes
+    badRequest2.proceed();
+    badRequest2.waitForCondition(() -> refreshCount.get() == 3, 2000);
+
+    // Allow the third bad request to complete
```
enocom, line 469: Is this comment correct?

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: Because I want to ensure that the retry is working - that the chain of futures is being built correctly even after the first failure. Early iterations of this PR had a bug where it would retry after one failure, but stop after the second failure.

core/src/test/java/com/google/cloud/sql/core/CloudSqlInstanceTest.java
hessjcg: fixed.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -165,43 +189,77 @@ String getPreferredIp(List<String> preferredTypes, long timeoutMs) {
...
+                  String.format(
+                      "[%s] Refresh Operation: Acquired rate limiter permit. Starting refresh...",
+                      instanceName));
+            },
+            executor);
+
+    // Once rate limiter is done, attempt to getInstanceData.
+    ListenableFuture<InstanceData> dataFuture =
+        Futures.whenAllComplete(rateLimit)
+            .callAsync(
```
enocom, line 242: Open question: why callAsync and not transformAsync?

enocom (approved):

enocom:
For my own reference, this is a recreation of https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/1457.

PR #1593, opened 2023-10-06
chore: Set project-wide config for CLIRR version compatibility checks

hessjcg:
Update the maven configuration so that the CLIRR plugin only checks classes that are documented
to be part of our public API.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -240,6 +240,23 @@
           </dependency>
         </dependencies>
       </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>clirr-maven-plugin</artifactId>
+        <configuration>
+          <includes>com/google/cloud/sql/CredentialFactory</includes>
+          <includes>com/google/cloud/sql/core/CoreSocketFactory</includes>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -240,6 +240,24 @@
...
       </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>clirr-maven-plugin</artifactId>
+        <configuration>
+          <includes>com/google/cloud/sql/CredentialFactory</includes>
+          <includes>com/google/cloud/sql/core/CoreSocketFactory</includes>
+          <includes>com/google/cloud/sql/mariadb/SocketFactory</includes>
+          <includes>com/google/cloud/sql/mariadb/SocketFactory</includes>
```
ttosta-google, line 251: This line is the same as the one above.

ttosta-google (approved):

pom.xml
hessjcg: Good catch. Fixed.

hessjcg:
Update the maven configuration so that the CLIRR plugin only checks classes that are documented
to be part of our public API.

pom.xml
ttosta-google: LGTM

Issue #992, opened 2022-09-21
Add a "lazy strategy" option that prevents refreshes from happening in the background


johanblumenberg:
## Bug Description

Every now and then we see a failure to refresh the ephemeral certificate used to connect to Cloud SQL.
Most of the time this is just an annoying error in the log, but the service tries again and succeeds most of the time. This is still annoying, since it adds a lot of noise to our monitoring.
But every now and then also the retry is failing, and we lose traffic because we don't have any certificate to authenticate to the DB.

The error that we get is this:

```
Got more than one input failure. Logging failures after the first
java.lang.RuntimeException: [...] Failed to update metadata for Cloud SQL instance.
	at com.google.cloud.sql.core.CloudSqlInstance.addExceptionContext(CloudSqlInstance.java:598)
	at com.google.cloud.sql.core.CloudSqlInstance.fetchMetadata(CloudSqlInstance.java:505)
	at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:131)
	at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:74)
	at com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:82)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.io.IOException: Error writing to server
	at java.base/sun.net.www.protocol.http.HttpURLConnection.writeRequests(HttpURLConnection.java:718)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.writeRequests(HttpURLConnection.java:730)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1613)
[...]

PR #1600, opened 2023-10-09
fix: Pause connection requests after certificate expires until refresh completes.

hessjcg:
When the application attempts to create a database connection, and the certificate or auth token has expired,
pause until the certificate and token are refreshed. If the token is not refreshed before the timeout, throw
an exception because the connection attempt failed.


enocom (commented):
I think this counts as a feature as we're changing how refresh is done.

I know we've talked about it, but can you make a succinct case why we need this change? It feels like an edge case for when we're in a CPU throttled context, which we need to address more directly with #992.

hessjcg:
Converting to draft until we decide to move forward.

Issue #776, opened 2022-03-14
Ability to recreate the coreSocketFactory in CoreSocketFactory


meltsufin:
Currently, `coreSocketFactory` is a singleton, and there is no way to recreate it after it has been created.

Needed here: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/982

kurtisvg:
The change needed for this will probably be somewhat involved as there are a bunch of assumptions in the current code that the CoreSocketFactory stays the same throughout.

meltsufin:
@kurtisvg Can you give an example of such an assumption? Thanks!

kurtisvg:
So looking at the code, it might not be as bad as I thought (I was originally thinking about replacing the Admin API client). I think the biggest problem is making sure the [executor is shutdown correctly](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/8a549cac72eae184703721e958d830404affc5fe/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L175) and that all of the instances in the hashmap are ["cleaned up" correctly](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/8a549cac72eae184703721e958d830404affc5fe/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L93) (and that the hashamp is reset)



We'll obviously need to make sure the method used to reset the instance is synchronized as well, and make sure none of the static methods rely on the codeSocketFactory directly and unsynchronized - they need to be using `getInstance` ([we may have a couple of examples of this already](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/8a549cac72eae184703721e958d830404affc5fe/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L448)).

enocom:
We're working through some significant refactoring in the core socket factory right now and will be able to revisit this request. As part of recreating the singleton, we'd be resetting the cached TLS certificates and basically starting again from scratch. Is that the expected behavior @meltsufin?

meltsufin:
Yes, having a public API to reset it is the requirement there.

hessjcg:
#1604 did not fix the whole issue.

PR #1604, opened 2023-10-10
feat: Java API to configure named connections

### LINKED ISSUE #1226: Allow callers to configure multiple "connectors" per socket factory
### Feature Description

This is a feature request to add support for a "Connector" type that can be registered with the SocketFactory.

Among other things, this will allow customers to configure a different set of credentials per connector.

The Connector API should look like this:

``` java
// Initialize a builder
JdbcConnectorBuilder builder = CoreSocketFactory.JdbcBuilder("my-connector"); // naming the connector

// Configure and register a connector
Connector myConnector = builder
  .withIAMAuthN() // defaults to password
  .withIPType("PUBLIC") // defaults to public,private
  .withCredentialFactory(myCredentialFactory) // using the credential factory interface
  .build(); 

// Register the connector programmatically
CoreSocketFactory.registerConnector(myConnector);

// OR
// Register the connector from Properties.
Properties properties = new Properties();
CoreSocketFactory.registerConnectorFromProperties(properties);

// Use the connector with a JDBC URL
String jdbcURL = "jdbc:mysql:///<DATABASE_NAME>" +
"?cloudSqlInstance=<INSTANCE_CONNECTION_NAME>" +
"&socketFactory=com.google.cloud.sql.mysql.SocketFactory"+
"&user=<MYSQL_USER_NAME>"+
"&password=<MYSQL_USER_PASSWORD>"+
"&cloudSqlConnectorName=my-connector" // <-- uses all the settings with the named connector

// Close a connector to stop its background refresh. Future attempted uses will throw an exception.
CoreSocketFactory.close("my-connector")

// Close all connectors
CoreSocketFactory.closeAll()
```

By default if callers don't register a connector, we will hash the URL and use that as the name.

### Sample code

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

hessjcg:
In real usage, an application would first call `Connector.register()` to register the configuration for a named connection.

```java
ConnectorConfig config =  new ConnectorConfig.Builder()
            .withTargetPrincipal("example@project.iam.googleapis.com")
            .withDelegates(Arrays.asList("delegate@project.iam.googleapis.com"))
            .build();
ConnectorRegistry.register("my-connector", config);
```

Then the application would use the connector by adding the `cloudSqlNamedConnector` parameter to the JDBC URL: 

```java
String jdbcUrl = "jdbc:mysql:///<DATABASE_NAME>?"+
  + "cloudSqlInstance=project:region:instance"
  + "&cloudSqlNamedConnector=my-connector"
  + "&socketFactory=com.google.cloud.sql.mysql.SocketFactory"
  + "&user=<DB_USER>&password=<PASSWORD>";
```

Or using JDBC connection properties, by adding the `cloudSqlNamedConnector` property: 

```java
// Set up URL parameters
Properties connProps=new Properties();
connProps.setProperty("user","<DB_USER>");
connProps.setProperty("password","<PASSWORD>");
connProps.setProperty("sslmode","disable");
connProps.setProperty("socketFactory","<DRIVER_CLASS>");
connProps.setProperty("cloudSqlInstance","my-connection");
connProps.setProperty("cloudSqlNamedConnector","[my-connection](project:region:instance)");

// Initialize connection pool
HikariConfig config=new HikariConfig();
config.setJdbcUrl("jdbc:mysql:///<DB_NAME>");
config.setDataSourceProperties(connProps);
config.setConnectionTimeout(10000); // 10s

HikariDataSource connectionPool=new HikariDataSource(config);
```

The application can stop individual named connectors, or all connectors: 
```java
ConnectorRegistry.close("my-connector");
```

The application can shutdown the connector, preventing any new connections
and stoping all connector threads. 

```java
ConnectorRegistry.shutdown();
```

Fixes #1226

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
+++ b/core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql;
+
+import com.google.cloud.sql.core.InternalConnectorRegistry;
+
+/** Configure the CloudSQL JDBC Connector. */
+public final class ConnectorRegistry {
+
+  /**
+   * Register a named connection so that it can later be referenced by name in a JDBC or R2DBC URL.
+   *
+   * @param name the named connection name.
+   * @param config the full configuration of the connection.
+   */
+  public static void register(String name, ConnectorConfig config) {
+    InternalConnectorRegistry.getInstance().register(name, config);
+  }
+
+  /**
+   * Close a named connector. This will stop all background credential refresh processes. All future
+   * attempts to connect via this named connection will fail.
+   *
+   * @param name the name of the connector to close.
+   */
+  public static void close(String name) {
+    InternalConnectorRegistry.getInstance().close(name);
+  }
+
+  /**
+   * Shutdown the entire CloudSQL JDBC Connector. This will stop all background threads. All future
+   * attempts to connect to a CloudSQL database will fail.
+   */
+  public static void shutdown() {
+    InternalConnectorRegistry.shutdownInstance();
+  }
+
+  /**
+   * Adds an external application name to the user agent string for tracking. This is known to be
+   * used by the spring-cloud-gcp project.
+   *
+   * @throws IllegalStateException if the SQLAdmin client has already been initialized
+   */
+  public static void addArtifactId(String artifactId) {
+    InternalConnectorRegistry.getInstance().addArtifactId(artifactId);
+  }
+}
--- a/core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
+++ b/core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
@@ -56,8 +56,9 @@ public final class InternalConnectorRegistry {
   private static final long MIN_REFRESH_DELAY_MS = 30000; // Minimum 30 seconds between refresh.
   private static InternalConnectorRegistry internalConnectorRegistry;
   private final ListenableFuture<KeyPair> localKeyPair;
-  private final ConcurrentHashMap<ConnectorConfig, Connector> connectors =
+  private final ConcurrentHashMap<ConnectorConfig, Connector> unnamedConnectors =
       new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, Connector> namedConnectors = new ConcurrentHashMap<>();
   private final ListeningScheduledExecutorService executor;
   private final CredentialFactory credentialFactory;
   private final int serverProxyPort;
@@ -109,6 +110,18 @@ public final class InternalConnectorRegistry {
     return internalConnectorRegistry;
   }
 
+  /**
+   * Calls shutdown on the singleton and removes the singleton. After calling shutdownInstance(),
+   * the next call to getInstance() will start a new singleton instance.
+   */
+  public static synchronized void shutdownInstance() {
+    if (internalConnectorRegistry != null) {
+      InternalConnectorRegistry old = internalConnectorRegistry;
+      internalConnectorRegistry = null;
+      old.shutdown();
+    }
+  }
+
   // TODO(kvg): Figure out better executor to use for testing
   @VisibleForTesting
   // Returns a listenable, scheduled executor that exits upon shutdown.
@@ -128,15 +141,20 @@ public final class InternalConnectorRegistry {
   }
 
   /**
-   * Creates a socket representing a connection to a Cloud SQL instance.
+   * Internal use only: Creates a socket representing a connection to a Cloud SQL instance.
    *
    * <p>Depending on the given properties, it may return either a SSL Socket or a Unix Socket.
    *
-   * @param config Configuration used to configure the connection.
+   * @param config used to configure the connection.
    * @return the newly created Socket.
    * @throws IOException if error occurs during socket creation.
    */
   public Socket connect(ConnectionConfig config) throws IOException, InterruptedException {
+    if (config.getNamedConnector() != null) {
+      Connector connector = getNamedConnector(config.getNamedConnector());
+      return connector.connect(config.withConnectorConfig(connector.getConfig()));
+    }
+
     // Validate parameters
     Preconditions.checkArgument(
         config.getCloudSqlInstance() != null,
@@ -146,7 +164,15 @@ public final class InternalConnectorRegistry {
     return getConnector(config).connect(config);
   }
 
-  public ConnectionMetadata getConnectionMetadata(ConnectionConfig config) throws IOException {
+  /** Internal use only: Returns ConnectionMetadata for a connection. */
+  public ConnectionMetadata getConnectionMetadata(ConnectionConfig config) {
+    if (config.getNamedConnector() != null) {
+      Connector connector = getNamedConnector(config.getNamedConnector());
+      return connector
+          .getConnection(config.withConnectorConfig(connector.getConfig()))
+          .getConnectionMetadata(refreshTimeoutMs);
+    }
+
     return getConnector(config).getConnection(config).getConnectionMetadata(refreshTimeoutMs);
   }
 
@@ -228,7 +254,7 @@ public final class InternalConnectorRegistry {
   }
 
   private Connector getConnector(ConnectionConfig config) {
-    return connectors.computeIfAbsent(
+    return unnamedConnectors.computeIfAbsent(
         config.getConnectorConfig(), k -> createConnector(config.getConnectorConfig()));
   }
 
@@ -255,6 +281,7 @@ public final class InternalConnectorRegistry {
         connectionInfoRepositoryFactory.create(credential, config);
 
     return new Connector(
+        config,
         adminApi,
         instanceCredentialFactory,
         executor,
@@ -262,5 +289,39 @@ public final class InternalConnectorRegistry {
         MIN_REFRESH_DELAY_MS,
         refreshTimeoutMs,
         serverProxyPort);
+  }
+
+  /** Register the configuration for a named connector. */
+  public void register(String name, ConnectorConfig config) {
+    if (this.namedConnectors.containsKey(name)) {
+      throw new IllegalArgumentException("Named connection " + name + " exists.");
+    }
+    this.namedConnectors.put(name, createConnector(config));
+  }
+
+  /** Close a named connector, stopping the refresh process and removing it from the registry. */
+  public void close(String name) {
+    Connector connector = namedConnectors.remove(name);
+    if (connector == null) {
+      throw new IllegalArgumentException("Named connection " + name + " does not exist.");
+    }
+    connector.close();
+  }
+
+  /** Shutdown all connectors and remove the singleton instance. */
+  public void shutdown() {
+    this.unnamedConnectors.forEach((key, c) -> c.close());
+    this.unnamedConnectors.clear();
+    this.namedConnectors.forEach((key, c) -> c.close());
+    this.namedConnectors.clear();
+    this.executor.shutdown();
+  }
+
+  private Connector getNamedConnector(String name) {
+    Connector connector = namedConnectors.get(name);
+    if (connector == null) {
+      throw new IllegalArgumentException("Named connection " + name + " does not exist.");
+    }
+    return connector;
   }
 }
--- a/core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
+++ b/core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
@@ -34,6 +34,7 @@ import java.net.Socket;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
+import java.util.Properties;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -313,12 +314,127 @@ public class InternalConnectorRegistryTest extends CloudSqlCoreTestingBase {
         () -> InternalConnectorRegistry.setApplicationName("sample-app"));
   }
 
+  @Test
+  public void registerConnection() throws IOException, InterruptedException {
+
+    String adminRootUrl = "https://googleapis.example.com/";
+    String adminServicePath = "sqladmin/";
+    String baseUrl = adminRootUrl + adminServicePath;
+
+    // The mock AdminApi only responds to requests where the URL starts with a custom admin api
+    // root url: https://googleapis.example.com/sqladmin/
+    InternalConnectorRegistry registry = createRegistry(PUBLIC_IP, credentialFactory, baseUrl);
+
+    // Register a ConnectionConfig named "my-connection" that uses the custom admin api root url.
+    ConnectorConfig configWithDetails =
+        new ConnectorConfig.Builder()
+            .withAdminRootUrl(adminRootUrl)
+            .withAdminServicePath(adminServicePath)
+            .build();
+    registry.register("my-connection", configWithDetails);
+
+    // Assert that when the named connection config is used, the socket opens correctly.
+    Properties goodProps = new Properties();
+    goodProps.setProperty(ConnectionConfig.CLOUD_SQL_NAMED_CONNECTOR_PROPERTY, "my-connection");
+    goodProps.setProperty(
+        ConnectionConfig.CLOUD_SQL_INSTANCE_PROPERTY, "myProject:myRegion:myInstance");
+    ConnectionConfig goodConfig = ConnectionConfig.fromConnectionProperties(goodProps);
+
+    Socket socket = registry.connect(goodConfig);
+    assertThat(readLine(socket)).isEqualTo(SERVER_MESSAGE);
+
+    // Assert that when the named connection config is not used, the socket fails to open.
+    Properties badProps = new Properties();
+    badProps.setProperty(
+        ConnectionConfig.CLOUD_SQL_INSTANCE_PROPERTY, "myProject:myRegion:myInstance");
+    ConnectionConfig badConfig = ConnectionConfig.fromConnectionProperties(badProps);
+
+    assertThrows(RuntimeException.class, () -> registry.connect(badConfig));
+  }
+
+  @Test
+  public void registerConnectionFailsWithDuplicateName() throws InterruptedException {
+    InternalConnectorRegistry registry = createRegistry(PUBLIC_IP, credentialFactory);
+    // Register a ConnectionConfig named "my-connection"
+    ConnectorConfig configWithDetails = new ConnectorConfig.Builder().build();
+    registry.register("my-connection", configWithDetails);
+
+    // Assert that you can't register a connection with a duplicate name
+    assertThrows(
+        IllegalArgumentException.class,
+        () -> registry.register("my-connection", configWithDetails));
+  }
+
+  @Test
+  public void registerConnectionFailsWithDuplicateNameAndDifferentConfig()
+      throws InterruptedException {
+    InternalConnectorRegistry registry = createRegistry(PUBLIC_IP, credentialFactory);
+
+    ConnectorConfig config =
+        new ConnectorConfig.Builder().withTargetPrincipal("joe@test.com").build();
+    registry.register("my-connection", config);
+
+    ConnectorConfig config2 =
+        new ConnectorConfig.Builder().withTargetPrincipal("jane@test.com").build();
+
+    // Assert that you can't register a connection with a duplicate name
+    assertThrows(IllegalArgumentException.class, () -> registry.register("my-connection", config2));
+  }
+
+  @Test
+  public void closeNamedConnectionFailsWhenNotFound() throws InterruptedException {
+    InternalConnectorRegistry registry = createRegistry(PUBLIC_IP, credentialFactory);
+    // Assert that you can't close a connection that doesn't exist
+    assertThrows(IllegalArgumentException.class, () -> registry.close("my-connection"));
+  }
+
+  @Test
+  public void connectFailsOnClosedNamedConnection() throws InterruptedException {
+    InternalConnectorRegistry registry = createRegistry(PUBLIC_IP, credentialFactory);
+    // Register a ConnectionConfig named "my-connection"
+    ConnectorConfig configWithDetails = new ConnectorConfig.Builder().build();
+    registry.register("my-connection", configWithDetails);
+
+    // Close the named connection.
+    registry.close("my-connection");
+
+    // Attempt and fail to connect using the cloudSqlNamedConnection connection property
+    Properties connProps = new Properties();
+    connProps.setProperty(ConnectionConfig.CLOUD_SQL_NAMED_CONNECTOR_PROPERTY, "my-connection");
+    ConnectionConfig nameOnlyConfig = ConnectionConfig.fromConnectionProperties(connProps);
+
+    // Assert that no connection is possible because the connector is closed.
+    IllegalArgumentException ex =
+        assertThrows(IllegalArgumentException.class, () -> registry.connect(nameOnlyConfig));
+    assertThat(ex).hasMessageThat().contains("Named connection my-connection does not exist.");
+  }
+
+  @Test
+  public void connectFailsOnUnknownNamedConnection() throws InterruptedException {
+    InternalConnectorRegistry registry = createRegistry(PUBLIC_IP, credentialFactory);
+
+    // Attempt and fail to connect using the cloudSqlNamedConnection connection property
+    Properties connProps = new Properties();
+    connProps.setProperty(ConnectionConfig.CLOUD_SQL_NAMED_CONNECTOR_PROPERTY, "my-connection");
+    ConnectionConfig nameOnlyConfig = ConnectionConfig.fromConnectionProperties(connProps);
+    IllegalArgumentException ex =
+        assertThrows(IllegalArgumentException.class, () -> registry.connect(nameOnlyConfig));
+    assertThat(ex).hasMessageThat().contains("Named connection my-connection does not exist.");
+  }
+
   private InternalConnectorRegistry createRegistry(
       String ipType, CredentialFactory credentialFactory) throws InterruptedException {
+    return createRegistry(ipType, credentialFactory, null);
+  }
+
+  private InternalConnectorRegistry createRegistry(
+      String ipType, CredentialFactory credentialFactory, String baseUrl)
+      throws InterruptedException {
     FakeSslServer sslServer = new FakeSslServer();
     int port = sslServer.start(ipType);
     ConnectionInfoRepositoryFactory factory =
-        new StubConnectionInfoRepositoryFactory(fakeSuccessHttpTransport(Duration.ofSeconds(0)));
+        new StubConnectionInfoRepositoryFactory(
+            fakeSuccessHttpTransport(Duration.ofSeconds(0), baseUrl));
     return new InternalConnectorRegistry(
         clientKeyPair, factory, credentialFactory, port, TEST_MAX_REFRESH_MS, defaultExecutor);
   }
--- a/core/src/main/java/com/google/cloud/sql/core/ConnectionConfig.java
+++ b/core/src/main/java/com/google/cloud/sql/core/ConnectionConfig.java
@@ -167,6 +167,18 @@ public class ConnectionConfig {
     this.authType = authType;
   }
 
+  /** Creates a new instance of the ConnectionConfig with an updated connectorConfig. */
+  public ConnectionConfig withConnectorConfig(ConnectorConfig config) {
+    return new ConnectionConfig(
+        cloudSqlInstance,
+        namedConnector,
+        unixSocketPath,
+        ipTypes,
+        authType,
+        unixSocketPathSuffix,
+        config);
+  }
+
   public String getNamedConnector() {
     return namedConnector;
   }
--- a/core/src/main/java/com/google/cloud/sql/core/Connector.java
+++ b/core/src/main/java/com/google/cloud/sql/core/Connector.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.sql.core;
 
+import com.google.cloud.sql.ConnectorConfig;
 import com.google.cloud.sql.CredentialFactory;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningScheduledExecutorService;
@@ -43,8 +44,10 @@ class Connector {
       new ConcurrentHashMap<>();
   private final long refreshTimeoutMs;
   private final int serverProxyPort;
+  private final ConnectorConfig config;
 
   Connector(
+      ConnectorConfig config,
       DefaultConnectionInfoRepository adminApi,
       CredentialFactory instanceCredentialFactory,
       ListeningScheduledExecutorService executor,
@@ -52,6 +55,7 @@ class Connector {
       long minRefreshDelayMs,
       long refreshTimeoutMs,
       int serverProxyPort) {
+    this.config = config;
     this.adminApi = adminApi;
     this.instanceCredentialFactory = instanceCredentialFactory;
     this.executor = executor;
@@ -59,6 +63,10 @@ class Connector {
     this.minRefreshDelayMs = minRefreshDelayMs;
     this.refreshTimeoutMs = refreshTimeoutMs;
     this.serverProxyPort = serverProxyPort;
+  }
+
+  public ConnectorConfig getConfig() {
+    return config;
   }
 
   /** Extracts the Unix socket argument from specified properties object. If unset, returns null. */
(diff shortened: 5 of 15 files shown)


core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
```diff
@@ -364,6 +364,50 @@ public void testGetApplicationNameFailsAfterInitialization() {
         IllegalStateException.class, () -> CoreSocketFactory.setApplicationName("sample-app"));
   }
 
+  @Test
+  public void create_successfulNamedConnection() throws IOException, InterruptedException {
```
hessjcg, line 368: These tests show how to use the `Connector` API, since CoreSocketFactory is the implementation of `Connector`

In real usage, an application would first call 
```
 ConnectionConfig real =
        new ConnectionConfig.Builder()
            .withCloudSqlInstance("myProject:myRegion:myInstance")
            .withIpTypes("PRIMARY")
            .build();
ConnectorRegistry.getConnector().register("my-connection", config);
```

And then to use the connection, they would use the JDBC URL: 
```
jdbc:mysql:///<DATABASE_NAME>?cloudSqlInstanceName=my-connection&socketFactory=com.google.cloud.sql.mysql.SocketFactory&user=<MYSQL_USER_NAME>&password=<MYSQL_USER_PASSWORD>
```

enocom:
``` java
ConnectionConfig real =  new ConnectionConfig.Builder()
            .withCloudSqlInstance("myProject:myRegion:myInstance")
            .withIpTypes("PRIMARY")
            .build();
ConnectorRegistry.getConnector().register("my-connection", config);
```

I'm not sure this will work with multiple credentials. A configuration object should correspond 1:1 with a connector object.

Instead it should be:

``` java
ConnectionConfig config =  new ConnectionConfig.Builder()
            .withCloudSqlInstance("myProject:myRegion:myInstance")
            .withIpTypes("PRIMARY")
            .withName("my-connection")
            .build();

// Maybe we rename CoreSocketFactory to ConnectorRegistry, but we can move ahead
// with CoreSocketFactory for now
CoreSocketFacotry.getInstance().register(config);


// which under the hood does this where registeredConnectors is a ConcurrentHashMap
this.registeredConnectors.set(config.getName(), new Connector( /*... credentials, ip type etc ...*/));

// and in client code:

Properties connProps = new Properties();
connProps.setProperty("cloudSqlInstanceName", "my-connection");

// and in connect

Connector connector = this.registeredConnectors.get(properties.get("cloudSqlInstanceName"));
connector.connect(); // ... etc

```

core/src/main/java/com/google/cloud/sql/Connector.java
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql;
+
+/** Manage named connector configurations, leveraging the JDBC URL. */
+public interface Connector {
```
enocom, line 20: What is the purpose of this interface? Do we need an interface when only one class implements it? Also why is it public? It's not part of our public API itself.

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql;
+
+import com.google.cloud.sql.core.CoreSocketFactory;
+
+/** The public java API to configure the connector registry. */
+public enum ConnectorRegistry {
```
enocom, line 22: The name is nice, but maybe we should just make CoreSocketFactory.getInstance() public?

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
```diff
@@ -0,0 +1,34 @@
...
+  INSTANCE;
+
+  /** Returns the singleton instance of the Connector. */
+  public static Connector getConnector() {
```
enocom, line 26: This is not a "connector" it's a `getInstance` or `getRegistry`.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
```diff
@@ -332,4 +345,25 @@ ListenableFuture<InstanceData> getCurrent() {
   public CloudSqlInstanceName getInstanceName() {
     return instanceName;
   }
+
+  public void close() {
```
enocom, line 349: Should this be public? Shouldn't the close be exposed on the CoreSocketFactory instead?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -394,4 +413,54 @@ private CloudSqlInstance apiFetcher(ConnectionConfig config) {
...
+            "No named connection named " + config.getNamedConnection());
+      }
+      return namedConfig;
+    }
+    return config;
+  }
+
+  @Override
+  public void register(String name, ConnectionConfig config) {
+    ConnectionConfig oldConfig = this.namedConfigs.putIfAbsent(name, config);
```
enocom, line 431: We should probably throw an exception if two registrations occur for the same name.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -394,4 +413,54 @@ private CloudSqlInstance apiFetcher(ConnectionConfig config) {
...
+    if (oldConfig != null) {
+      CloudSqlInstance oldInstance = this.namedInstances.get(name);
+      if (oldInstance != null) {
+        oldInstance.close();
+      }
+    }
+  }
+
+  @Override
+  public void close(String name) {
```
enocom, line 441: Same thing here: closing an unknown connector should throw an exception.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -394,4 +413,54 @@ private CloudSqlInstance apiFetcher(ConnectionConfig config) {
...
+    this.namedInstances.remove(name);
+    CloudSqlInstance oldInstance = this.namedInstances.remove(name);
+    if (oldInstance != null) {
+      oldInstance.close();
+    }
+    this.namedConfigs.remove(name);
+  }
+
+  @Override
+  public void shutdown() {
```
enocom, line 451: Does this implementation lend itself to restarting the CoreSocketFactory? See https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/776 for example.

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
```diff
@@ -364,6 +365,67 @@ public void testGetApplicationNameFailsAfterInitialization() {
...
+
+    FakeSslServer sslServer = new FakeSslServer();
+    int port = sslServer.start(PUBLIC_IP);
+    ApiFetcherFactory factory =
+        new StubApiFetcherFactory(fakeSuccessHttpTransport(Duration.ofSeconds(0)));
+
+    CoreSocketFactory registry =
```
enocom, line 376: Let's be consistent with names and types. Unfortunately, we did not name this ConnectorRegistry, and so should stick with what we have.

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
enocom: Alternatively, we could do this (which seems more keeping with the existing interface):

``` java
CoreSocketFactory.register("my-connection", config);
```

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -111,6 +112,10 @@ public final class CoreSocketFactory {
   private static CoreSocketFactory coreSocketFactory;
   private final ListenableFuture<KeyPair> localKeyPair;
   private final ConcurrentHashMap<String, CloudSqlInstance> instances = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, ConnectionConfig> namedConfigs =
```
enocom, line 115: Can we achieve this functionality with only one ConcurrentHashMap?

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
hessjcg: No, that's a separate use case. For #776, we will want to create a `restart()` method that essentially does something similar, but doesn't shutdown() the executor.

core/src/test/java/com/google/cloud/sql/core/CoreSocketFactoryTest.java
hessjcg: Fixed. the test variables are now named `coreSocketFactory`.

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java
hessjcg: Fixed. This should be package scope.

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
hessjcg: This matches Alloy DB JDBC Connector's [ConnectorRegistry.getConnector()](https://github.com/GoogleCloudPlatform/alloydb-java-connector/blob/8496eecee4787a6e912973bcbe4bfa2255732c5b/alloydb-jdbc-connector/src/main/java/com/google/cloud/alloydb/ConnectorRegistry.java#L62C4-L64C4)

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
hessjcg: Let's discuss some design options.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
hessjcg: This separates the named instances from the unnamed instances. Separate maps is the cleanest way to avoid unintentional conflicts in the key spaces. 

Also, we have been debating whether to change the key type for unnamed instances to include more properties than just the instance name. That implementation would not use the String as the key for unnamed instances.

core/src/main/java/com/google/cloud/sql/Connector.java
hessjcg: Let's discuss.

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
enocom: My vote is for:

```
CoreSocketFactory.register("my-connection", config);
```

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
enocom: The interface, yes. But the CoreSocketFactory is more like AlloyDB java's ConnectorRegistry. I think we should just treat the CoreSocketFactory as an equivalent to the registry and not introduce a wrapper class here.

ttosta-google (commented):
Please update `ConnectionConfigTest` to test the new property `cloudSqlNamedConnection`.

Also we should add a new integration test for this feature.

hessjcg:
@ttosta-google Good catch. I've added assertions to the test.

hessjcg:
Converting back to Draft until we finish some renaming and other refactorying.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -57,6 +57,9 @@ public final class InternalConnectorRegistry {
   private static InternalConnectorRegistry internalConnectorRegistry;
   private final ListenableFuture<KeyPair> localKeyPair;
   private final ConcurrentHashMap<ConnectorKey, Connector> connectors = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, ConnectionConfig> namedConfigs =
+      new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, Connector> namedConnectors = new ConcurrentHashMap<>();
```
enocom, line 62: Could we update ConnectorKey to account for connector name and thereby combine this map with the `connectors` map above?

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
```diff
@@ -367,6 +368,107 @@ public void testGetApplicationNameFailsAfterInitialization() {
...
+  public void registerConnection() throws IOException, InterruptedException {
+    FakeSslServer sslServer = new FakeSslServer();
+    int port = sslServer.start(PUBLIC_IP);
+    ConnectionInfoRepositoryFactory factory =
+        new StubConnectionInfoRepositoryFactory(fakeSuccessHttpTransport(Duration.ofSeconds(0)));
+    InternalConnectorRegistry coreSocketFactory =
+        new InternalConnectorRegistry(
+            clientKeyPair, factory, credentialFactory, port, TEST_MAX_REFRESH_MS, defaultExecutor);
+    // Register a ConnectionConfig named "my-connection"
+    ConnectionConfig real =
```
enocom, line 381: nit: s/real/connectionConfig/

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
```diff
@@ -367,6 +368,107 @@ public void testGetApplicationNameFailsAfterInitialization() {
...
+            .withIpTypes("PRIMARY")
+            .build();
+    coreSocketFactory.register("my-connection", real);
+    // Attempt to connect using the cloudSqlNamedConnection connection property
+    Properties connProps = new Properties();
+    connProps.setProperty(ConnectionConfig.CLOUD_SQL_NAMED_CONNECTION_PROPERTY, "my-connection");
+    ConnectionConfig namedConfig = ConnectionConfig.fromConnectionProperties(connProps);
+    Socket socket = coreSocketFactory.connect(namedConfig);
+    // Assert that the socket opens correctly.
+    assertThat(readLine(socket)).isEqualTo(SERVER_MESSAGE);
```
enocom, line 393: How does this assertion verify the registered config was used? Could we simplify this assertion to make that point more clear?

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
```diff
@@ -367,6 +368,107 @@ public void testGetApplicationNameFailsAfterInitialization() {
         () -> InternalConnectorRegistry.setApplicationName("sample-app"));
   }
 
+  @Test
+  public void registerConnection() throws IOException, InterruptedException {
+    FakeSslServer sslServer = new FakeSslServer();
```
enocom, line 373: There's a lot of duplicate setup here. Could we consolidate it to make these tests easier to read?

enocom (changes requested):
This is looking really good.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -146,6 +151,7 @@ public Socket connect(ConnectionConfig config) throws IOException, InterruptedEx
   }
 
   public ConnectionMetadata getConnectionMetadata(ConnectionConfig config) throws IOException {
+    config = resolveConfig(config);
```
ttosta-google, line 154: Is it required here?

`resolveConfig()` is called inside `getConnector()` as well.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +272,53 @@ private Connector createConnector(ConnectionConfig config) {
...
+            "No named connection named " + config.getNamedConnection());
+      }
+      return namedConfig;
+    }
+    return config;
+  }
+
+  /** Register the configuration for a named connector. */
+  public void register(String name, ConnectionConfig config) {
+    ConnectionConfig oldConfig = this.namedConfigs.putIfAbsent(name, config);
```
ttosta-google, line 290: What if we create the connector instead of just saving the ConnectionConfig in the Hashmap? This way we don't need `namedConfigs` and when `getConnector()` is called it returns the connector from `namedConnectors` or `connectors`.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +272,53 @@ private Connector createConnector(ConnectionConfig config) {
...
+    if (oldConfig != null) {
+      throw new IllegalArgumentException("Named configuration " + name + " already exists.");
+    }
+  }
+
+  /** Close a named connector, stopping the refresh process and removing it from the registry. */
+  public void close(String name) {
+    this.namedConnectors.remove(name);
```
ttosta-google, line 298: `namedConnectors.remove(name)` is being called twice.

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
```diff
@@ -367,6 +310,102 @@ public void testGetApplicationNameFailsAfterInitialization() {
...
+
+  @Test
+  public void closeNamedConnectionFailsWhenNotFound() throws InterruptedException {
+    InternalConnectorRegistry registry = createRegistry(PUBLIC_IP, credentialFactory);
+    // Assert that you can't close a connection that doesn't exist
+    assertThrows(IllegalArgumentException.class, () -> registry.close("my-connection"));
+  }
+
+  @Test
+  public void closeNamedConnection() throws InterruptedException {
```
ttosta-google, line 362: maybe `connectFailsOnClosedNamedConnection` ?

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql;
+
+import com.google.cloud.sql.core.InternalConnectorRegistry;
+
+/** The public java api to control the Connector configuration. */
+public enum ConnectorRegistry {
```
ttosta-google, line 22: In AlloyDB the `ConnectorRegistry` implements `Closeable` and overrides:

```
@Override
  public void close() {}
```

Should we do it here so shutdown all the connectors in case the user does not do it.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
enocom: +1 -- I was thinking about this problem and came to the same thought. If we just create a map of connectors, we can avoid this problem of multiple concurrent hashmaps.

enocom (changes requested):
Let's see if we can store connectors in a map as @ttosta-google suggests. Also, make sure to update the PR comment in this that describes the API. It's currently stale and doesn't accurately describe the API added here.

Otherwise, I think we're almost there with this work. :clap:

hessjcg:
I'll update ([jdbc.md](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/jdbc.md) and [r2dbc.md](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/r2dbc.md)) in a separate PR to document the new public API.

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+  INSTANCE;
```
enocom, line 23: Does this need to be a singleton since we're call through to a singleton?

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
         refreshTimeoutMs,
         serverProxyPort);
   }
+
+  private ConnectionConfig resolveConfig(ConnectionConfig config) {
```
enocom, line 271: Could we encapsulate this into ConnectionConfig? Something like `connectionConfig.getKey()` and then do this:

```
return connectors.computeIfAbsent(config.getKey(), k -> createConnector(config));
```

hessjcg: No, because `ConnectionConfig` is in the public API package and `ConnectionKey` is in the internal package. We want `ConnectionKey` to only be used internally, and to change as needed.

core/src/main/java/com/google/cloud/sql/ConnectorRegistry.java
hessjcg: Enums must have at least 1 value. Instead, I will switch this to `final class` and get rid of `INSTANCE`.

hessjcg (commented):
I have updated this PR: 

- I updated InternalConnectorRegistry to use a single map to hold Connector instances. This change required ~70 additional lines of code, and I believe that it made the code harder to understand.
- I changed the ConnectorRegistry public methods to static methods.
- I added Javadoc to the ConnectorRegistry public methods.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
...
+    if (config.getNamedConnection() != null) {
+      ConnectorKey namedConfig = resolveKey(config);
```
ttosta-google, line 273: rename to `connectorKey` or something similar

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
...
+      if (namedConfig == null) {
+        throw new IllegalArgumentException(
+            "Named configuration " + config.getNamedConnection() + " does not exist.");
```
ttosta-google, line 276: This message is not clear. Maybe "Named connection "%" does not exist."

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
...
+      }
+      return namedConfig.getConfig();
+    }
+    return config;
+  }
+
+  private ConnectorKey resolveKey(ConnectionConfig config) {
+    Optional<ConnectorKey> namedConfig =
```
ttosta-google, line 284: rename to `connectorKey` or something similar

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
...
+            .findFirst();
+    return namedConfig.orElse(null);
+  }
+
+  /** Register the configuration for a named connector. */
+  public void register(String name, ConnectionConfig config) {
+    ConnectionConfig resolvedConfig = config.withNamedConnection(name);
+    ConnectorKey key = new ConnectorKey(resolvedConfig);
+    if (this.connectors.containsKey(key)) {
+      throw new IllegalArgumentException("Named configuration " + name + " exists.");
```
ttosta-google, line 296: This message is not clear. Maybe "Named Connection "%" already exists."

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
...
+    }
+    this.connectors.put(key, createConnector(resolvedConfig));
+  }
+
+  /** Close a named connector, stopping the refresh process and removing it from the registry. */
+  public void close(String name) {
+    ConnectionConfig namedConfig = new ConnectionConfig.Builder().withNamedConnection(name).build();
+    ConnectorKey oldKey = resolveKey(namedConfig);
+    if (oldKey == null) {
+      throw new IllegalArgumentException("Named configuration " + name + " does not exist.");
```
ttosta-google, line 306: Same here

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
...
+      if (namedConfig == null) {
+        throw new IllegalArgumentException(
+            "Named configuration " + config.getNamedConnection() + " does not exist.");
+      }
+      return namedConfig.getConfig();
+    }
+    return config;
+  }
+
+  private ConnectorKey resolveKey(ConnectionConfig config) {
```
ttosta-google, line 283: What if it receives only the "String namedConnection" ? There is no need for the entire `ConnectionConfig`.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -261,4 +267,55 @@ private Connector createConnector(ConnectionConfig config) {
...
+            .filter(k -> config.getNamedConnection().equals(k.getConfig().getNamedConnection()))
+            .findFirst();
+    return namedConfig.orElse(null);
+  }
+
+  /** Register the configuration for a named connector. */
+  public void register(String name, ConnectionConfig config) {
+    ConnectionConfig resolvedConfig = config.withNamedConnection(name);
+    ConnectorKey key = new ConnectorKey(resolvedConfig);
+    if (this.connectors.containsKey(key)) {
```
ttosta-google, line 295: To verify if this name is already being used we should not consider the entire ConnectionConfig. 

The way it is implemented, the following will not throw an exception:

```
ConnectionConfig config =  new ConnectionConfig.Builder()
            .withCloudSqlInstance("myProject:myRegion:myInstance")
            .withIpTypes("PRIMARY")
            .build();
ConnectorRegistry.getConnector().register("my-connection", config);

ConnectionConfig config2 =  new ConnectionConfig.Builder()
            .withCloudSqlInstance("myProject:myRegion:myInstance")
            .withIpTypes("PRIVATE")
            .build();
ConnectorRegistry.getConnector().register("my-connection", config2);
```

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: I've updated this and also `connect()` to hopefully be more straightforward.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Good idea. Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Great catch. This is a bug. I didn't see this when I rewrote the connector map behavior. 

To get the correct behavior the InternalConnectorRegistry will need to store the ConnectionConfig and Connector by name. This is why I originally implemented this with two maps: a map of Connector with "name" as a key, and a map of ConnectionConfig with name as a key. 

Maybe we need to revisit whether "one map for all connectors" will get us the behavior we want.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Good catch. Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Fixed.

hessjcg (commented):
@ttosta-google I fixed a few errors. You did catch a good bug. Can we discuss how best to proceed?

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: I have added a new test: `InternalConnectorRegistryTest.registerConnectionFailsWithDuplicateNameAndDifferentConfig()`

hessjcg: I updated the implementation of `InternalConnectorRegistry` so that it has two maps again and the new test will now pass. Let me know what you think.

hessjcg:
Converting back to draft. This got too big so I split it into several PRs: 

- #1671 
- #1672 
- #1673

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -263,4 +290,39 @@ private Connector createConnector(ConnectorConfig config) {
...
+    if (connector == null) {
+      throw new IllegalArgumentException("Named connection " + name + " does not exist.");
+    }
+    connector.close();
+  }
+
+  /** Shutdown all connectors and remove the singleton instance. */
+  public void shutdown() {
+    this.unnamedConnectors.forEach(
+        (key, c) -> {
```
enocom, line 314: nit: could this be all in one line?

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -263,4 +290,39 @@ private Connector createConnector(ConnectorConfig config) {
...
+          c.close();
+        });
+    this.unnamedConnectors.clear();
```
enocom, line 317: We need to close all the unnamed connectors in the `connectors` map.

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
enocom: Sorry I still don't understand -- how do we know the registered connector was used and not simply an unnamed connector?

enocom (changes requested):
Two things and then we can merge this:

- Let's make sure to close all unnamed connectors
- The tests of named connectors aren't clear in how they're confirming we're using a named connector.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -127,26 +142,40 @@ static ListeningScheduledExecutorService getDefaultExecutor() {
...
   public Socket connect(ConnectionConfig config) throws IOException, InterruptedException {
-    // Validate parameters
-    Preconditions.checkArgument(
-        config.getCloudSqlInstance() != null,
-        "cloudSqlInstance property not set. Please specify this property in the JDBC URL or the "
-            + "connection Properties with value in form \"project:region:instance\"");
+    if (config.getNamedConnection() != null) {
+      NamedConnector namedConnector = getNamedConnector(config);
+      return namedConnector.connector.connect(namedConnector.config);
+    } else {
```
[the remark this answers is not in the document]
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -128,26 +141,39 @@ static ListeningScheduledExecutorService getDefaultExecutor() {
...
-  public ConnectionMetadata getConnectionMetadata(ConnectionConfig config) throws IOException {
-    return getConnector(config).getConnection(config).getConnectionMetadata(refreshTimeoutMs);
+  /** Internal use only: Returns ConnectionMetadata for a connection. */
+  public ConnectionMetadata getConnectionMetadata(ConnectionConfig config) {
+    if (config.getNamedConnector() != null) {
+      Connector connector = getNamedConnector(config.getNamedConnector());
+      return connector
+          .getConnection(config.withConnectorConfig(connector.getConfig()))
+          .getConnectionMetadata(refreshTimeoutMs);
+    } else {
```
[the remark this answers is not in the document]
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: Fixed.

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
hessjcg: After the most recent change to ConnectorConfig (from ConnectionConfig), this test doesn't actually test anything anymore. 

I've updated this to clearly test that the named connector is used.

enocom (approved):

PR #1615, opened 2023-10-13
chore: Extract Refresher from CloudSqlInstance

hessjcg:
Create a new class Refresher which knows how to refresh a data object with an expiration date. 
This moves the refresh algorithm out of CloudSqlInstance so that it is easier to share the refresh algorithm 
implementation between this project and the AlloyDB Java Connector.  

The api contract of `CloudSqlInstance` used by `CoreSocketFactory` does not change. `CloudSqlInstance` now has a private `Refresher` which it configures to keep `InstanceData` valid. 

```
refresher =
    new Refresher<>(
        connectionName, // The name of the connector, in refresh log messages
        executor, // The executor
        // Supplier<ListenableFuture<InstanceData>> which retrieves fresh data.
        () ->  instanceDataSupplier.getInstanceData(
                      this.instanceName, this.accessTokenSupplier, this.authType, executor, keyPair),
        new RateLimitCalculator(minRefreshDelayMs)); // The rate limiter.
```

Another benefit: by encapsulating the refresh algorithm encapsulated in a separate class, it will be much easier to keep the AlloyDB Java Connector and the Cloud SQL Java Connector refresh implementations in sync. Without this refactor, we will need to copy the logic from CloudSqlInstance and paste it into DefaultConnectionInfoCache. Then, we will need to make A LOT of subtle changes. It will be very hard to keep these implementations in sync over time.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/InstanceData.java
+++ b/core/src/main/java/com/google/cloud/sql/core/InstanceData.java
@@ -36,7 +36,7 @@ class InstanceData {
     this.expiration = expiration;
   }
 
-  Instant getExpiration() {
+  public Instant getExpiration() {
     return expiration;
   }
 
--- a/core/src/main/java/com/google/cloud/sql/core/Refresher.java
+++ b/core/src/main/java/com/google/cloud/sql/core/Refresher.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import com.google.common.base.Throwables;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningScheduledExecutorService;
+import com.google.errorprone.annotations.concurrent.GuardedBy;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Handles periodic refresh operations for an instance. */
+class Refresher {
+  private static final Logger logger = Logger.getLogger(CloudSqlInstance.class.getName());
+
+  private final ListeningScheduledExecutorService executor;
+
+  private final Object instanceDataGuard = new Object();
+  private final AsyncRateLimiter rateLimiter;
+
+  private final RefreshCalculator refreshCalculator = new RefreshCalculator();
+  private final Supplier<ListenableFuture<InstanceData>> refreshOperation;
+  private final String name;
+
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/google/cloud/sql/core/ExpiringData.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import java.time.Instant;
+
+/** A piece of data that has an expiration time. */
+public interface ExpiringData {
```
enocom, line 22: Does this need to be public?

core/src/main/java/com/google/cloud/sql/core/Refresher.java
```diff
@@ -0,0 +1,248 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Handles periodic refresh operations for an instance. */
+public class Refresher<T extends ExpiringData> {
```
enocom, line 34: Same here. Public?

enocom (commented):
This looks like a nice improvement.

Two things:

1. Let's be judicious about our use of "public"
2. Do we need the interface for ExpiringData? If so, why?

core/src/main/java/com/google/cloud/sql/core/Refresher.java
hessjcg: Good point. This is now package scope.

core/src/main/java/com/google/cloud/sql/core/ExpiringData.java
hessjcg: Good point. This is now package scope.

core/src/main/java/com/google/cloud/sql/core/InstanceData.java
```diff
@@ -22,7 +22,7 @@
 import javax.net.ssl.SSLContext;
 
 /** Represents the results of a certificate and metadata refresh operation. */
-class InstanceData {
+class InstanceData implements ExpiringData {
```
enocom, line 25: Why do we think this interface is a useful addition?

core/src/main/java/com/google/cloud/sql/core/Refresher.java
```diff
@@ -0,0 +1,248 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Handles periodic refresh operations for an instance. */
+class Refresher<T extends ExpiringData> {
```
enocom, line 34: Does this need to be generic? We'll only ever be refreshing "InstanceData" aka "ConnectionInfo".

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/Refresher.java
hessjcg: For now, yes. We want to copy/paste this into AlloyDB unchanged.  Once we're using the same name for ConnectionInfo in both projects, we can eliminate the type parameter. 

Also, in the future I can see us using this for other things too - like refreshing just the IAM Authn credentials for an implementation of HikariPool that supports "direct path iam authn" connections.

core/src/main/java/com/google/cloud/sql/core/InstanceData.java
hessjcg: This enables the type parameter on Refresher.

core/src/main/java/com/google/cloud/sql/core/Refresher.java
enocom: Copy/paste + search and replace is cheap. I think we want to avoid introducing this abstraction until we need it.

enocom (approved):

PR #1605, opened 2023-10-10
docs: Consolidate all JDBC docs onto one page

hessjcg:
This change consolidates all the JDBC documentation from the 4 different files into a single file. This will make it
much easier to maintain and expand the documentation. 

Part of #1550

PR #1617, opened 2023-10-16
feat: add support for configuring Admin API URL

### LINKED ISSUE #1587: Add support for configuring Admin API URL
Currently we have a private property that supports configuring the SQL Admin API URL. We should add this property to the supported properties in the connection string.

Add two new connection properties: "cloudSqlAdminRootUrl" and "cloudSqlAdminServicePath". These names use the common prefix "cloudSql" and also match the SQLAdmin java api naming scheme for "rootUrl" and "servicePath". 

`SqlAdminApiFetcher.create()` configures the `SQLAdmin` client library for each connection.  Update this method
so that when the admin api and root url are set, 

```
SQLAdmin.Builder adminApiBuilder; // ...
adminApiBuilder.setRootUrl(rootUrl);
adminApiBuilder.setServicePath(servicePath);
```

Related to https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/1575.

ttosta-google:
Fixes #1587

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/ConnectionConfig.java
+++ b/core/src/main/java/com/google/cloud/sql/ConnectionConfig.java
@@ -220,6 +220,7 @@ public class ConnectionConfig {
       return this;
     }
 
+    /** Builds a new instance of {@code ConnectionConfig}. */
     public ConnectionConfig build() {
       return new ConnectionConfig(
           cloudSqlInstance,
--- a/core/src/main/java/com/google/cloud/sql/ConnectionConfig.java
+++ b/core/src/main/java/com/google/cloud/sql/ConnectionConfig.java
@@ -32,6 +32,8 @@ public class ConnectionConfig {
   public static final String CLOUD_SQL_INSTANCE_PROPERTY = "cloudSqlInstance";
   public static final String CLOUD_SQL_DELEGATES_PROPERTY = "cloudSqlDelegates";
   public static final String CLOUD_SQL_TARGET_PRINCIPAL_PROPERTY = "cloudSqlTargetPrincipal";
+  public static final String CLOUD_SQL_ADMIN_ROOT_URL_PROPERTY = "cloudSqlAdminRootUrl";
+  public static final String CLOUD_SQL_ADMIN_SERVICE_PATH_PROPERTY = "cloudSqlAdminServicePath";
   public static final String UNIX_SOCKET_PROPERTY = "unixSocketPath";
   public static final String ENABLE_IAM_AUTH_PROPERTY = "enableIamAuth";
   public static final String IP_TYPES_PROPERTY = "ipTypes";
@@ -45,6 +47,8 @@ public class ConnectionConfig {
   private final String unixSocketPath;
   private final AuthType authType;
   private final List<IpType> ipTypes;
+  private final String adminRootUrl;
+  private final String adminServicePath;
 
   /** Create a new ConnectionConfig from the well known JDBC Connection properties. */
   public static ConnectionConfig fromConnectionProperties(Properties props) {
@@ -67,9 +71,20 @@ public class ConnectionConfig {
         listIpTypes(
             props.getProperty(
                 ConnectionConfig.IP_TYPES_PROPERTY, ConnectionConfig.DEFAULT_IP_TYPES));
+    final String adminRootUrl =
+        props.getProperty(ConnectionConfig.CLOUD_SQL_ADMIN_ROOT_URL_PROPERTY);
+    final String adminServicePath =
+        props.getProperty(ConnectionConfig.CLOUD_SQL_ADMIN_SERVICE_PATH_PROPERTY);
 
     return new ConnectionConfig(
-        csqlInstanceName, targetPrincipal, delegates, unixSocketPath, authType, ipTypes);
+        csqlInstanceName,
+        targetPrincipal,
+        delegates,
+        unixSocketPath,
+        authType,
+        ipTypes,
+        adminRootUrl,
+        adminServicePath);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


ttosta-google:
Note: I will update **docs/jdbc.md** once the PR #1605 is merged.

core/src/test/java/com/google/cloud/sql/core/MockAdminApi.java
```diff
@@ -66,10 +66,10 @@ public class MockAdminApi {
 
   private static final Pattern CONNECT_SETTINGS_PATTERN =
       Pattern.compile(
-          ".*/sql/v1beta4/projects/(?<project>.*)/instances/(?<instance>.*)/connectSettings");
+          "(?<baseUrl>.*)sql/v1beta4/projects/(?<project>.*)/instances/(?<instance>.*)/connectSettings");
   private static final Pattern GENERATE_EPHEMERAL_CERT_PATTERN =
       Pattern.compile(
-          ".*/sql/v1beta4/projects/(?<project>.*)/instances/(?<instance>.*):generateEphemeralCert");
+          "(?<baseUrl>.*)sql/v1beta4/projects/(?<project>.*)/instances/(?<instance>.*):generateEphemeralCert");
```
hessjcg, line 72: Nice. This is as good as an integration test, since it actually exercises the http client.

core/src/main/java/com/google/cloud/sql/ConnectionConfig.java
```diff
@@ -134,6 +153,14 @@ public List<IpType> getIpTypes() {
     return ipTypes;
   }
 
+  public String getAdminRootUrl() {
+    return adminRootUrl;
+  }
+
+  public String getlAdminServicePath() {
```
hessjcg, line 160: I think this is a typo. Maybe it should be `getAdminServicePath`

docs/r2dbc.md
```diff
@@ -268,6 +268,24 @@ The environment's application default principal impersonates
...
+The Java Connector supports setting the SQL Admin API URL with the
+`ADMIN_ROOT_URL` and `ADMIN_SERVICE_PATH` options. The `ADMIN_ROOT_URL`
+option specifies the URL-encoded root URL of the service and the
+`ADMIN_SERVICE_PATH` option specifies the URL-encoded service path of the
+service.
+
+### Example
+
+```java
+ConnectionFactoryOptions options=ConnectionFactoryOptions.builder()
```
hessjcg, line 282: More realistic example strings for the docs: 
"https://googleapis.example.com/"
"sqladmin/v1/"

Also, we may want to direct people to see the javadoc for `setRootUrl` and `setServicePath`: https://cloud.google.com/java/docs/reference/google-api-client/latest/com.google.api.client.googleapis.services.AbstractGoogleClient.Builder#com_google_api_client_googleapis_services_AbstractGoogleClient_Builder_setRootUrl_java_lang_String_

enocom (commented):
Do we want to keep service path? I believe we don't use that in any meaningful way in this project or any other connectors.

docs/r2dbc.md
```diff
@@ -268,6 +268,24 @@ The environment's application default principal impersonates
...
 impersonates the TARGET_SERVICE_ACCOUNT.
 
+## SQL Admin API URL
+
+The Java Connector supports setting the SQL Admin API URL with the
+`ADMIN_ROOT_URL` and `ADMIN_SERVICE_PATH` options. The `ADMIN_ROOT_URL`
+option specifies the URL-encoded root URL of the service and the
+`ADMIN_SERVICE_PATH` option specifies the URL-encoded service path of the
+service.
+
```
hessjcg, line 278: Please add a few more pieces of information here: 
- Explain why someone may use this: "Some applications need to connect to a Google Cloud API other than the GCP public API."
- Copy the detailed rules from the underlying api docs:`https://cloud.google.com/java/docs/reference/google-api-client/latest/com.google.api.client.googleapis.services.AbstractGoogleClient.Builder#com_google_api_client_googleapis_services_AbstractGoogleClient_Builder_setRootUrl_java_lang_String_`
- This is optional, describe the default value. "If this is not set, the connector will use the public Google Cloud API." 

The default values from the [underlying driver class](https://github.com/googleapis/google-api-java-client-services/blob/7346122e3aea1d088018a17830e55839f4e6a8bf/clients/google-api-services-sqladmin/v1beta4/2.0.0/com/google/api/services/sqladmin/SQLAdmin.java#L61C27-L61C27) are 

```
DEFAULT_ROOT_URL = "https://sqladmin.googleapis.com/"
DEFAULT_SERVICE_PATH = ""
```

Also, here are some more realistic values for the example. I think `tasks/v1/` will confuse people.
```
    .option(ADMIN_ROOT_URL, "https://googleapis.example.com/");
    .option(ADMIN_SERVICE_PATH, "sqladmin/")
```

hessjcg: Finally, in a separate PR we will need to add this to jdbc.md, after #1605 is merged.

hessjcg (changes requested):
I've requested some minor changes to the documentation. The implementation looks great! Nice work.

docs/r2dbc.md
ttosta-google: Updated.

core/src/main/java/com/google/cloud/sql/ConnectionConfig.java
ttosta-google: Thank you for catching that. Updated.

ttosta-google:
> Do we want to keep service path? I believe we don't use that in any meaningful way in this project or any other connectors.

I don't have a preference. I could remove it. @hessjcg what do you think?

enocom:
For historical reasons we've always used v1beta4, even though there's a v1. This is to ensure v1beta4 continues to work, given the significant customer usage.

docs/r2dbc.md
ttosta-google: Updated **jdbc.md**

hessjcg: Nice.

hessjcg (approved):
LGTM. Nice work!

PR #1618, opened 2023-10-16
refactor: Use IpType instead of String

hessjcg:
The new public configuration class IpType should be used throughtout our connector instead of using
strings to represent IpTypes. 

This moves the translation from a String returned by the API into the `SqlAdminApiFetcher`, the class that is responsible for translating API responses into internal data. 

The old unit test `SqlAdminApiFetcherTest.testFetchInstanceData_returnsIpAddresses()` confirms that the translation works correctly.


ttosta-google (approved):
LGTM

PR #1642, opened 2023-10-24
chore: rename InstanceData to ConnectionInfo

hessjcg:


core/src/main/java/com/google/cloud/sql/core/Refresher.java
```diff
@@ -89,8 +89,8 @@ class Refresher {
    * successful attempt. If no attempts succeed within the timeout, throws a RuntimeException with
    * the exception from the last failed refresh attempt as the cause.
    */
-  InstanceData getData(long timeoutMs) {
-    ListenableFuture<InstanceData> instanceDataFuture;
+  ConnectionInfo getData(long timeoutMs) {
+    ListenableFuture<ConnectionInfo> instanceDataFuture;
```
enocom, line 93: Let's make sure to grab all the field names too.

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/Refresher.java
hessjcg: Fixed

enocom (approved):

PR #1645, opened 2023-10-25
chore: Extract internal connector registry implementation

hessjcg:
Move the connector registry implementation out of CoreSocketFactory and into InternalConnectorRegistry. 

CoreSocketFactory is deprecated and contains only the static public methods used informally by other projects. 
It delegates to the InternalConnectorRegistry.


ttosta-google (approved):
LGTM

enocom (changes requested):
Let's frame this as "extract internal-only registry" as we're leaving CoreSocketFactory in place.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
```diff
@@ -0,0 +1,361 @@
...
+/**
+ * Factory responsible for obtaining an ephemeral certificate, if necessary, and establishing a
+ * secure connecting to a Cloud SQL instance.
+ *
+ * <p>This class should not be used directly, but only through the JDBC driver specific {@code
+ * SocketFactory} implementations.
+ *
+ * <p>The API of this class is subject to change without notice.
+ */
+public final class CloudSqlConnectorRegistry {
```
enocom, line 55: Shall we remove all the public references in here and make this package protected before merging this?

enocom, line 55: Shall we call this "Internal" something or other? I worry the comment isn't enough to deter usage.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
```diff
@@ -0,0 +1,361 @@
...
+import java.util.logging.Logger;
+import javax.net.ssl.SSLSocket;
+import jnr.unixsocket.UnixSocketAddress;
+import jnr.unixsocket.UnixSocketChannel;
+
+/**
+ * Factory responsible for obtaining an ephemeral certificate, if necessary, and establishing a
+ * secure connecting to a Cloud SQL instance.
+ *
+ * <p>This class should not be used directly, but only through the JDBC driver specific {@code
```
enocom, line 50: Let's make it crystal clear: "Internal class made public for module usage only. This is not part of the official public API and should not be used directly." Or something similar.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -16,41 +16,11 @@
...
 
 /**
- * Factory responsible for obtaining an ephemeral certificate, if necessary, and establishing a
- * secure connecting to a Cloud SQL instance.
- *
- * <p>This class should not be used directly, but only through the JDBC driver specific {@code
- * SocketFactory} implementations.
- *
- * <p>The API of this class is subject to change without notice.
+ * Implementation of informally used java API to preserve compatibility with older code that uses
```
enocom, line 22: Will we be marking this as a deprecated class in a separate PR?

core/src/test/java/com/google/cloud/sql/core/CloudSqlConnectorRegistryTest.java
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright 2016 Google Inc.
+ * Copyright 2023 Google LLC
```
enocom, line 2: These can stay as is, since we're only moving the file.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
enocom: Perhaps with a "WARNING" or similar call-out to lead in.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
enocom: My vote would be for InternalConnectorRegistry, PrivateConnectorRegistry, or siimilar.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
```diff
@@ -0,0 +1,361 @@
...
+    this.connectionInfoRepositoryFactory = connectionInfoRepositoryFactory;
+    this.credentialFactory = credentialFactory;
+    this.serverProxyPort = serverProxyPort;
+    this.executor = executor;
+    this.localKeyPair = localKeyPair;
+    this.refreshTimeoutMs = refreshTimeoutMs;
+  }
+
+  /** Returns the {@link CloudSqlConnectorRegistry} singleton. */
+  public static synchronized CloudSqlConnectorRegistry getInstance() {
```
enocom, line 99: Does this need to be public? I know we're going to turn this into an enum as we have in AlloyDB, but shall we keep the API surface as small as possible during that refactor?

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
```diff
@@ -0,0 +1,361 @@
...
+    return System.getProperty(USER_TOKEN_PROPERTY_NAME, "");
+  }
+
+  /**
+   * Adds an external application name to the user agent string for tracking. This is known to be
+   * used by the spring-cloud-gcp project.
+   *
+   * @throws IllegalStateException if the SQLAdmin client has already been initialized
+   */
+  public static void setApplicationName(String applicationName) {
```
enocom, line 285: What shall we do about this? I forget what the plan here was.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
hessjcg: I like InternalConnectorRegistry best.

core/src/test/java/com/google/cloud/sql/core/CloudSqlConnectorRegistryTest.java
hessjcg: Sadly, since this is a new file, the header-check requires this year's date.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
hessjcg: This needs to be public so that it can be accessed by `com.google.cloud.sql.ConnectorRegistry`, which is in a different package. We can look into restricting external access to the `core` package using Java Modules some day.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
hessjcg: We're going to deprecate this method in favor of `ConnectorRegistry.addApplicationId(String id)`.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
hessjcg: I have now annotated this class with `@Deprecated`.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
enocom: @hessjcg let's add some more warnings to the doc string.

core/src/main/java/com/google/cloud/sql/core/CloudSqlConnectorRegistry.java
enocom: OK that's fine. Let's make it clear that this is a "private" method in that case with a similar warning.

enocom (commented):
LGTM

I'd like to see some serious warnings around our public methods on the InternalConnectorRegistry before we merge this.

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -16,42 +16,16 @@
...
- * secure connecting to a Cloud SQL instance.
+ * WARNING: This class is not a stable, public Java API. The class definition may change without
+ * notice. Implementation of informally used java API to preserve compatibility with older code that
+ * uses CoreSocketFactory.
  *
- * <p>This class should not be used directly, but only through the JDBC driver specific {@code
- * SocketFactory} implementations.
- *
- * <p>The API of this class is subject to change without notice.
+ * @deprecated This will soon be replaced.
```
enocom, line 26: Replaced with what? Can we be more specific so if someone sees this they'll know what to look for and do instead?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -98,303 +72,11 @@ public final class CoreSocketFactory {
...
-                ConnectionConfig.CLOUD_SQL_TARGET_PRINCIPAL_PROPERTY,
-                ConnectionConfig.CLOUD_SQL_DELEGATES_PROPERTY));
-      }
-      instanceCredentialFactory = credentialFactory;
-    }
-
-    HttpRequestInitializer credential = instanceCredentialFactory.create();
-    DefaultConnectionInfoRepository adminApi =
-        connectionInfoRepositoryFactory.create(credential, config);
+  @Deprecated
```
enocom, line 75: Should we mention what people should do instead?

core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java
```diff
@@ -16,42 +16,16 @@
...
-import java.util.concurrent.ScheduledThreadPoolExecutor;
-import java.util.logging.Logger;
-import javax.net.ssl.SSLSocket;
-import jnr.unixsocket.UnixSocketAddress;
-import jnr.unixsocket.UnixSocketChannel;
 
 /**
- * Factory responsible for obtaining an ephemeral certificate, if necessary, and establishing a
- * secure connecting to a Cloud SQL instance.
+ * WARNING: This class is not a stable, public Java API. The class definition may change without
```
enocom, line 22: I think we put this in the wrong place -- this needs to be on the InternalConnectorRegistry.

The bit about compatibility should stay.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -0,0 +1,361 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.logging.Logger;
+import javax.net.ssl.SSLSocket;
+import jnr.unixsocket.UnixSocketAddress;
+import jnr.unixsocket.UnixSocketChannel;
+
+/**
+ * Factory responsible for obtaining an ephemeral certificate, if necessary, and establishing a
```
enocom, line 47: This isn't a factory. It's a holder of registered connectors. All the connection related code will move into a Connector in future commits.

core/src/main/java/com/google/cloud/sql/core/package-info.java
```diff
@@ -15,8 +15,10 @@
  */
 
 /**
- * Package com.google.cloud.sql.core holds internal shared packages that implement logic to create a
- * socket to a Cloud SQL database. Classes in this package are considered internal and subject to
+ * WARNING: This package does not contain any stable, public Java API. The class definitions may
  * change without notice.
+ *
+ * <p>Package com.google.cloud.sql.core holds internal shared packages that implement logic to
```
enocom, line 21: Let's add a note about com.google.cloud.sql as holding all the public API.

enocom (approved):

PR #1648, opened 2023-10-27
chore: Introduce ConnectionMetadata for R2DBC factories

hessjcg:
Make a single internal public method to be used by R2DBC drivers to get the metadata needed
to make a connection to a Cloud SQL instance: the TLS config, and the IP address.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/ConnectionMetadata.java
+++ b/core/src/main/java/com/google/cloud/sql/core/ConnectionMetadata.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.TrustManagerFactory;
+
+/**
+ * A value object containing configuration needed to set up an mTLS connection to a Cloud SQL
+ * instance.
+ */
+public class ConnectionMetadata {
+  private final String preferredIpAddress;
+  private final KeyManagerFactory keyManagerFactory;
+  private final TrustManagerFactory trustManagerFactory;
+
+  /** Construct an immutable ConnectionMetadata. */
+  public ConnectionMetadata(
+      String preferredIpAddress,
+      KeyManagerFactory keyManagerFactory,
+      TrustManagerFactory trustManagerFactory) {
+
+    this.preferredIpAddress = preferredIpAddress;
+    this.keyManagerFactory = keyManagerFactory;
+    this.trustManagerFactory = trustManagerFactory;
+  }
+
+  public String getPreferredIpAddress() {
+    return preferredIpAddress;
+  }
+
+  public KeyManagerFactory getKeyManagerFactory() {
+    return keyManagerFactory;
+  }
+
+  public TrustManagerFactory getTrustManagerFactory() {
+    return trustManagerFactory;
+  }
+}
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
@@ -24,8 +24,6 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningScheduledExecutorService;
 import java.io.IOException;
 import java.security.KeyPair;
-import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.net.ssl.SSLSocket;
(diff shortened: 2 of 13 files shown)


core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
```diff
@@ -107,6 +104,23 @@ SSLSocket createSslSocket(long timeoutMs) throws IOException {
         getConnectionInfo(timeoutMs).getSslContext().getSocketFactory().createSocket();
   }
 
+  ConnectionMetadata getConnectionMetadata(long timeoutMs) {
```
ttosta-google, line 107: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/core/Connector.java#L110 could call `instance.getConnectionMetadata(..).getIpAddress()` and the method `getPreferredIp()` could be deleted.

core/src/main/java/com/google/cloud/sql/core/ConnectionMetadata.java
```diff
@@ -0,0 +1,53 @@
...
+package com.google.cloud.sql.core;
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.TrustManagerFactory;
+
+/**
+ * A value object containing configuration needed to set up an mTLS connection to a Cloud SQL
+ * instance.
+ */
+public class ConnectionMetadata {
```
ttosta-google, line 26: We already have a `Metadata` class used by `ConnectionInfo`. Is it possible to merge them and have only one? 

So `ConnectionInfo` would have `ConnectionMetadata` instead of having both `Metadata` and `SslData`.

The use would be: 

```
ConnectionMetadata getConnectionMetadata(long timeoutMs) {
    return getConnectionInfo(timeoutMs).getConnectionMetadata();
}
```

hessjcg: I'm open to ideas. This is pretty messy. 

The best solution I could think of is to make the value object `ConnectionMetadata` that extracts only the essential info out these 3 classes plus ConnectionConfig hide this mess from the R2DBC driver. 

Unfortunately, I don't think we can easily combine `ConnectionMetadata`, `Metadata`, `SslData` and `ConnectionInfo` because they are created at different times and build on each other. 

(See DefaultConnectionInfoRepository)

Metadata, SslData, and ConnectionInfo are in separate classes because they get built at different phases of the refresh process. 

First two requests API requests are made simultaneously: 
- Metadata created from the GET InstanceMetadata API request. 
- an ephemereral certificate is created for the instance

Then when both are completed, an SslData is created from the combination of Metadata and ephemeral certificate. 

Finally, a ConnectionInfo is created from the Metadata and SslData.

core/src/main/java/com/google/cloud/sql/core/ConnectionMetadata.java
```diff
@@ -0,0 +1,53 @@
...
+  private final String ipAddress;
+  private final KeyManagerFactory keyManagerFactory;
+  private final TrustManagerFactory trustManagerFactory;
+
+  /** Construct an immutable ConnectionMetadata. */
+  public ConnectionMetadata(
+      String ipAddress,
```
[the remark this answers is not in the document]
hessjcg: Yes. Fixed.

core/src/main/java/com/google/cloud/sql/core/ConnectionMetadata.java
```diff
@@ -0,0 +1,53 @@
...
+      KeyManagerFactory keyManagerFactory,
```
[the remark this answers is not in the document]
hessjcg: I'm trying to hide SslData and avoid introducing a new internal, public class.

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
hessjcg: Good catch. Fixed.

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
```diff
@@ -107,6 +104,23 @@ SSLSocket createSslSocket(long timeoutMs) throws IOException {
...
+    ConnectionInfo info = getConnectionInfo(timeoutMs);
+    String preferredIp = null;
+
+    for (IpType ipType : config.getIpTypes()) {
```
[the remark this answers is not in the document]
hessjcg: That's an oversight, good catch. Fixed.

core/src/test/java/com/google/cloud/sql/core/TestDataSupplier.java
```diff
@@ -41,9 +43,17 @@ class TestDataSupplier implements ConnectionInfoRepository {
                   IpType.PRIVATE, "10.10.10.10",
                   IpType.PSC, "abcde.12345.us-central1.sql.goog"),
               null),
-          new SslData(null, null, null),
+          new SslData(null, newKms(), null),
           Instant.now().plus(1, ChronoUnit.HOURS));
 
+  private static KeyManagerFactory newKms() {
```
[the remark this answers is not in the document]
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/ConnectionMetadata.java
ttosta-google: Got it. 

Would it be possible to rename `Metadata` to `InstanceMetadata` ? 

It would be clear that `ConnectionMetadata` is related to the connection and `InstanceMetadata` to the Cloud SQL instance.

It's just a suggestion.

hessjcg: Yeah, `InstanceMetadata` seems more clear.

ttosta-google (approved):
LGTM

PR #1654, opened 2023-10-31
feat: Identify the connector for a ConnectionConfig by it's unique configuration

hessjcg:
Part of #1226

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
+++ b/core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
@@ -56,7 +56,7 @@ public final class InternalConnectorRegistry {
   private static final long MIN_REFRESH_DELAY_MS = 30000; // Minimum 30 seconds between refresh.
   private static InternalConnectorRegistry internalConnectorRegistry;
   private final ListenableFuture<KeyPair> localKeyPair;
-  private final ConcurrentHashMap<String, Connector> connectors = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<ConnectorKey, Connector> connectors = new ConcurrentHashMap<>();
   private final ListeningScheduledExecutorService executor;
   private final CredentialFactory credentialFactory;
   private final int serverProxyPort;
@@ -241,7 +241,7 @@ public final class InternalConnectorRegistry {
   }
 
   private Connector getConnector(ConnectionConfig config) {
-    return connectors.computeIfAbsent(config.getCloudSqlInstance(), k -> createConnector(config));
+    return connectors.computeIfAbsent(new ConnectorKey(config), k -> createConnector(config));
   }
 
   private Connector createConnector(ConnectionConfig config) {
--- a/core/src/test/java/com/google/cloud/sql/core/ConnectorKeyTest.java
+++ b/core/src/test/java/com/google/cloud/sql/core/ConnectorKeyTest.java
@@ -1,0 +1,301 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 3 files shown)


core/src/test/java/com/google/cloud/sql/core/ConnectorKeyTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+import static com.google.common.truth.Truth.assertThat;
+
+import com.google.cloud.sql.AuthType;
+import com.google.cloud.sql.ConnectionConfig;
+import java.util.Arrays;
+import org.junit.Test;
+
+public class ConnectorKeyTest {
+
```
ttosta-google, line 27: Please add tests for the other ConnectionConfig attributes (e.g. ipType)

core/src/test/java/com/google/cloud/sql/core/ConnectorKeyTest.java
```diff
@@ -0,0 +1,243 @@
...
+    ConnectorKey k2 =
+        new ConnectorKey(
+            new ConnectionConfig.Builder().withCloudSqlInstance("proj:name:inst2").build());
+
+    assertThat(k1).isEqualTo(k2);
+    assertThat(k1.hashCode()).isEqualTo(k2.hashCode());
+  }
+
+  @Test
+  public void adminUrlNotEqual() {
```
ttosta-google, line 55: Rename to `testKeysNotEqual_withAdminUrlNotEqual` or something similar. Same for the other tests.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -241,7 +241,7 @@ public static void setApplicationName(String applicationName) {
   }
 
   private Connector getConnector(ConnectionConfig config) {
-    return connectors.computeIfAbsent(config.getCloudSqlInstance(), k -> createConnector(config));
+    return connectors.computeIfAbsent(new ConnectorKey(config), k -> createConnector(config));
```
enocom, line 244: In AlloyDB, we used a string to distinguish. Do we prefer one approach over another here?

https://github.com/GoogleCloudPlatform/alloydb-java-connector/blob/7fc16e7e10df096fc0dd226144e348ae4b5c3866/alloydb-jdbc-connector/src/main/java/com/google/cloud/alloydb/ConnectionConfig.java#L78-L91

cc @ttosta-google

core/src/test/java/com/google/cloud/sql/core/ConnectorKeyTest.java
hessjcg: Done. Also, with better test case names.

core/src/test/java/com/google/cloud/sql/core/ConnectorKeyTest.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: I prefer this approach. Now that we have introduced connection configuration parameters set per-connection configuration like 'targetPrincipal', the InternalConnectorRegistry will need to create a connector for each unique configuration of connector parameters. In this implementation, the InternalConnectorRegistry will automatically create exactly as many connectors as the application needs, usually just 1 connector.

ttosta-google: I like this approach as well. It is cleaner and easy to add new parameters if needed.

ttosta-google (approved):

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
enocom: @ttosta-google let's add this to AlloyDB in that case. Would you file an issue?

ttosta-google: Should we do it as part of https://github.com/GoogleCloudPlatform/alloydb-java-connector/issues/273?

PR #1675, opened 2023-11-07
feat: Set the GoogleCredentials to use on a connector

### LINKED ISSUE #1670: Socket factory option for path to google credentials file
### Feature Description

This socket factory works great with IDEs like IntelliJ and really improves the experience with Cloud SQL. Currently, Google credentials are sourced by `GoogleCredentials.getApplicationDefault` which looks in known locations and an environment variable.

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/core/ApplicationDefaultCredentialFactory.java#L40

But when using the IDE for multiple projects with different credentials, it can be tricky having to relogin, or to manipulate the environment variable during IDE startup. If there were a socket factory option that could be set to a path of a credentials file, then it would be possible to register each data source in the IDE with a separate credential file for smooth operation on multiple projects.

### Sample code

```Java
getCredentials() {

  if (connectionConfig.getCredentialsPath() != null) {
    return GoogleCredentials.fromStream(new FileInputStream(connectionConfig.getCredentialsPath());
  }

}
```

or for avoiding SPI modification

```Java
class ApplicationDefaultCredentialFactory implements CredentialFactory {
  static HttpRequestInitializer createWithCredentialsPath(String path) {
    // Extract common code in real PR
   GoogleCredentials credentials = GoogleCredentials.fromStream(new FileInputStream(path));
   if (credentials.createScopedRequired()) {
    credentials = credentials.createScoped(..);
    }
    return new HttpCredentialsAdapter(credentials);
  }
}

public final class InternalConnectorRegistry {
  private Connector createConnector(ConnectionConfig config) {
    HttpRequestInitializer credential;
    if (config.getCredentialsPath() != null) {
     credential = ApplicationDefaultCredentialFactory.createWithCredentialsPath(config.getCredentialsPath());
    } else {
      // existing
    }
}
```

### Alternatives Considered

_No response_

### Additional Details

`ConnectionConfig` would need to be passed into the credential factory, which could be an involved breaking change, though the config itself is accessible when creating credentials. Perhaps it would make sense to add a side-channel in `ApplicationDefaultCredentialsFactory`, `static HttpRequestInitializer createWithCredentialsPath(String path)` since the presence of the config would indicate the SPI isn't required.

https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java#L251

hessjcg:
Provides new configuration fields ConnectorConfig that allow the user to connectors use a specific GoogleCredential. users may set one of these three fields: 
- `Supplier<GoogleCredentials> googleCredentialsSupplier` - a supplier to provide GoogleCredentials instances to the connector. 
- `GoogleCredentials googleCredentials` - an instance of GoogleCredentials to use for all requests. 
- `String googleCredentialsPath` - path to file containing GoogleCredentials json. 

Adds a JDBC Connection Property `cloudSqlGoogleCredentialsPath` to set the googleCredentialsPath field.

Fixes #1670

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/ConnectorConfig.java
+++ b/core/src/main/java/com/google/cloud/sql/ConnectorConfig.java
@@ -160,6 +160,23 @@ public class ConnectorConfig {
 
     /** Builds a new instance of {@code ConnectionConfig}. */
     public ConnectorConfig build() {
+      // validate only one GoogleCredentials configuration field set
+      int googleCredsCount = 0;
+      if (googleCredentials != null) {
+        googleCredsCount++;
+      }
+      if (googleCredentialsPath != null) {
+        googleCredsCount++;
+      }
+      if (googleCredentialsSupplier != null) {
+        googleCredsCount++;
+      }
+      if (googleCredsCount > 1) {
+        throw new IllegalStateException(
+            "Invalid configuration, more than one GoogleCredentials field has a value "
+                + "(googleCredentials, googleCredentialsPath, googleCredentialsSupplier)");
+      }
+
       return new ConnectorConfig(
           targetPrincipal,
           delegates,
--- a/core/src/main/java/com/google/cloud/sql/ConnectorConfig.java
+++ b/core/src/main/java/com/google/cloud/sql/ConnectorConfig.java
@@ -16,8 +16,10 @@
 
 package com.google.cloud.sql;
 
+import com.google.auth.oauth2.GoogleCredentials;
 import com.google.common.base.Objects;
 import java.util.List;
+import java.util.function.Supplier;
 
 /**
  * ConnectorConfig is an immutable configuration value object that holds the entire configuration of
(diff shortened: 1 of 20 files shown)


core/src/main/java/com/google/cloud/sql/core/CredentialFactoryProvider.java
```diff
@@ -45,4 +46,40 @@ static CredentialFactory getCredentialFactory() {
     }
     return credentialFactory;
   }
+
+  static CredentialFactory getInstanceCredentialFactory(
+      CredentialFactory defaultCredentialFactory, ConnectorConfig config) {
+
+    CredentialFactory instanceCredentialFactory;
+    if (config.getGoogleCredentialsSupplier() != null) {
```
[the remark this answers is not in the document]
hessjcg: Yes. Good catch. I added this to #1676.

core/src/main/java/com/google/cloud/sql/core/CredentialFactoryProvider.java
```diff
@@ -45,4 +46,40 @@ static CredentialFactory getCredentialFactory() {
     }
     return credentialFactory;
   }
+
+  static CredentialFactory getInstanceCredentialFactory(
+      CredentialFactory defaultCredentialFactory, ConnectorConfig config) {
```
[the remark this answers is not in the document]
hessjcg: You are right that this is weird. I'll rework CredentialFactoryProvider a bit so that this makes more sense. We want to reuse the defaultCredentialFactory instance.

core/src/main/java/com/google/cloud/sql/ConnectorConfig.java
```diff
@@ -108,7 +160,14 @@ public Builder withAdminServicePath(String adminServicePath) {
 
     /** Builds a new instance of {@code ConnectionConfig}. */
     public ConnectorConfig build() {
-      return new ConnectorConfig(targetPrincipal, delegates, adminRootUrl, adminServicePath);
+      return new ConnectorConfig(
```
enocom, line 163: I think build should fail if multiple credentials have been set. WDYT?

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/ConnectorConfig.java
ttosta-google: +1

hessjcg: Good point. For named connectors, `ConnectorConfig.Builder.build()` will throw an IllegalStateException if more than one field is set. 

For unnamed connectors, it is only possible to set one of these fields, so there is no chance of creating a bad state.

ttosta-google (approved):
LGTM

enocom (approved):

PR #1676, opened 2023-11-08
docs: Add docs for connector configuration. Part of #1226 #1670.

hessjcg:


docs/configuration.md
```diff
@@ -0,0 +1,128 @@
+# Connector Configuration Reference
+
+## Configuring Named Connectors
+
+The application can programatically configure named connectors using the java
```
enocom, line 5: Let's be super nit-picky about our docs here and make sure we're capitalizing proper nouns (Java), putting spaces between our equals in our Java code (as if it had been formatted), and just generally giving this a very careful read-over with a copy-editor like attention.

Also can we lead with a section on why people might want to do this? For example, "Sometimes it's necessary to configure multiple credential sources, different SA impersonation paths, etc. This API allows for this advanced usage." Or something similar to frame the bigger picture.

enocom: We should also distinguish between ConnectorConfig vs ConnectionConfig (or whatever we expose) to help our users understand how the two differ.

docs/configuration.md
```diff
@@ -0,0 +1,128 @@
...
+    config.setJdbcUrl("jdbc:mysql:///<DB_NAME>");
+    config.setDataSourceProperties(connProps);
+    config.setConnectionTimeout(10000); // 10s
+
+    HikariDataSource connectionPool=new HikariDataSource(config);
+```
+
+When using a named connector, the JDBC connection will ignore all connector
+configuration properties in the JDBC connection properties, using the named
+connector configuration instead. See the full list
```
enocom, line 60: Should we log an error if people make this mistake? Warning: using a named connector causes the cloudsqlDelegates (or whatever prop) to be ignored

docs/configuration.md
```diff
@@ -0,0 +1,128 @@
...
+  provide
+  `GoogleCredentials` instances to the connector.
+- `GoogleCredentials googleCredentials` - an instance of GoogleCredentials to
+  use
+  for all requests.
+- `String googleCredentialsPath` - path to file containing GoogleCredentials
+  json. May be set with the JDBC Connection
+  Property `cloudSqlGoogleCredentialsPath`.
+
+Users should only set one field. If more than one field is set, the connector
```
enocom, line 100: I think we should throw an exception and just not support setting multiple credentials per connector.

docs/configuration.md
```diff
@@ -0,0 +1,128 @@
...
+| unixSocketPath              | UNIX_SOCKET         | The path to the unix socket to connect to the database, when using the java connector with the [Cloud SQL Auth Proxy](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/)                                                             |                  | `/var/db/my-db-instance`          |
+| enableIamAuth               | ENABLE_IAM_AUTH     | Enable IAM Authentication to authenticate to the database. Valid values: `true` - authenticate with the IAM principal,  `false` - authenticate with a database user and password.                                                          | false            | `true`                            |
+| ipTypes                     | IP_TYPES            | A comma-separated list of IP types, ordered by preference. Value values: `PUBLIC` - connect to the instance's public IP, `PRIVATE` - connect to the instances private IP, `PSC` - connect to the instance through Private Service Connect. | `PUBLIC,PRIVATE` | `PSC,PRIVATE,PUBLIC`              |
+
+### Connector Configuration
```
enocom, line 118: Let's start with Connector. Since the flow goes create the connector -> configure the connection -> connect.

docs/configuration.md
```diff
@@ -0,0 +1,128 @@
+# Connector Configuration Reference
+
+## Configuring Named Connectors
```
enocom, line 3: Let's add a section on "unnamed connectors" and how they work.

enocom (changes requested):

docs/configuration.md
hessjcg: Good idea. In progress.

docs/configuration.md
hessjcg: Will do. Once the subject matter is tight, I'll will ask for help with a thorough proof-read.

docs/configuration.md
hessjcg: I don't think we should log an error. Most people will ignore it.

docs/configuration.md
hessjcg: Fixed. The code and docs are updated to throw an exception.

docs/configuration.md
hessjcg: Fixed.

docs/configuration.md
```diff
@@ -0,0 +1,261 @@
+# Connector Configuration Reference
+
```
hessjcg, line 2: Eno, Do we have any existing documentation that describes the difference between "connector" and "connection?"

enocom: No -- it's only a detail here. Other languages make this easier to understand.

docs/configuration.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+In R2DBC, use the `NAMED_CONNECTOR` ConnectionFactory option.
+
+When using a named connector, the JDBC connection will ignore all connector
+configuration properties in the JDBC connection properties, using the named
+connector configuration instead. See the full list
+of [connector configuration properties](#connector-configuration) below.
+
+The
```
enocom, line 86: Nit -- let's wrap this at 80 characters like this:

```
The test [JdbcPostgresNamedConnectorIntegrationTest.java][connector-example] 
contains a full example showing how to use a named connector.

[connector-example]: ../jdbc/postgres/src/test/java/com/google/cloud/sql/postgres/JdbcPostgresNamedConnectorIntegrationTests.java
```

docs/configuration.md
```diff
@@ -0,0 +1,261 @@
...
+
+You can close a named connector by calling `ConnectorRegistry.close()`. This
+will stop the certificate refresh process for that connector. It will also
+cause all subsequent attempts connect using this named connector to fail.
+
+```java
+ConnectorRegistry.close("my-connector");
+```
+
+Existing open database `Connection` instances will continue work until they
```
enocom, line 101: Nit: we're not naming a class when we say "Connection" so let's just make this:

"Existing open database connections will continue to work ..."

docs/configuration.md
```diff
@@ -0,0 +1,261 @@
...
+are closed.
+
+### Updating a Named Connector's Configuration
+
+If you wish to update a named connector's configuration, the application
+first calls `ConnectorRegistry.close()` and then `ConnectorRegistry.register()`
+with the new configuration. This creates a new connector with the new
+credentials.
+
+Existing open database `Connection` instances will continue work until they
```
enocom, line 111: Same here -- Existing open database connections will ...

enocom (approved):
A few nits. Otherwise looks great. Thanks for writing this up so nicely.

PR #1680, opened 2023-11-09
feat: use SLF4J for logging

### LINKED ISSUE #1594: Use slf4j logging
Currently we are using jdk logging. We should use slf4j logging instead.

ttosta-google:
Fixes #1594

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -124,8 +124,8 @@
       <dependency>
         <groupId>commons-codec</groupId>
         <artifactId>commons-codec</artifactId>
-      <version>1.16.0</version>
-    </dependency>
+        <version>1.16.0</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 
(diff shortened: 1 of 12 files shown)


enocom (commented):
Should this be a "feature"?

core/pom.xml
```diff
@@ -285,6 +286,18 @@
...
 
+    <!-- Logging -->
+    <dependency> 
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency> 
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
```
enocom, line 297: Should this be a test only dependency? 

It looks like this is meant for deployment time configuration otherwise: https://www.slf4j.org/manual.html#swapping.

ttosta-google: I've updated the scope to `runtime`. I think we should have logs to debug issues outside of the test scope. Please let me know what you think.

enocom: Maybe I've got this wrong, but I'd expect we code to the slf4j-api and then allow applications to supply the "provider". Since we do want logs in test, we should make sure we have a provider in a test scope. Otherwise, we don't want to impose an implementation on our customers.

See https://www.slf4j.org/manual.html#libraries for details.

hessjcg: What if we remove `slf4j-jdk14` dependency, and then add a `<scope>test</scope>` dependency for logback-classic-1.4.6.jar. I prefer Logback to Java 1.4 logging because I find it is much easier to configure.

ttosta-google: I've added a `<scope>test</scope>` dependency for `logback-classic` as suggested.

Note: the last `logback-classic` version that supports Java 8 is `1.3.11`

enocom (approved):

PR #1688, opened 2023-11-13
feat: Add reset() and shutdown(). Fixes #1687.

### LINKED ISSUE #776: Ability to recreate the coreSocketFactory in CoreSocketFactory

### LINKED ISSUE #1687: Reset or Shutdown the ConnectorRegistry
### Feature Description

An application should use the public Java API to either shut down or reset the connector registry. Update the methods 
as follows: 

- `ConnectorRegistry.shutdown()` - Stops all threads, clears all config, and causes future attempts to connect will  fail. 
- `ConnectorRegistry.reset()` - Stops all threads, clears all config, future attempts to connect will start a new Connector.

An application would use `shutdown()` when it receives a callback to gracefully stop itself. The application would then have gracefully stop all its background processes. The application calls shutdown() so that if some all other background threads cannot start a database connection after ConnectorRegistry.shutdown() is called.

An application would use `reset()` when it needs to reload its configuration. This would gracefully close all Connector instances, causing them to reload on the next connection attempt.

### Sample code

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

hessjcg:
Fixes #1687
Fixes #776

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
+++ b/core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
@@ -55,6 +55,7 @@ public final class InternalConnectorRegistry {
   private static final String version = getVersion();
   private static final long MIN_REFRESH_DELAY_MS = 30000; // Minimum 30 seconds between refresh.
   private static InternalConnectorRegistry internalConnectorRegistry;
+  private static boolean shutdown = false;
   private final ListenableFuture<KeyPair> localKeyPair;
   private final ConcurrentHashMap<ConnectorConfig, Connector> unnamedConnectors =
       new ConcurrentHashMap<>();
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
```diff
@@ -114,14 +119,24 @@ public static synchronized InternalConnectorRegistry getInstance() {
    * Calls shutdown on the singleton and removes the singleton. After calling shutdownInstance(),
    * the next call to getInstance() will start a new singleton instance.
    */
-  public static synchronized void shutdownInstance() {
+  public static synchronized void resetInstance() {
```
enocom, line 122: This should be private right?

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/InternalConnectorRegistry.java
hessjcg: It should be public so that `ConnectorRegistry.shutdown()` can call it. `ConnectorRegistry` is in a different package.

enocom (approved):

Issue #1558, opened 2023-09-28
Recover gracefully from laptop sleep


enocom:
Right now if a laptop goes to sleep with an active connection to a Cloud SQL instance, the socket factory won't recover on its own once the laptop has been restored.

There are limitations in Java to handling this:

https://stackoverflow.com/questions/52876556/how-does-java-calculate-sleep-time-when-pc-goes-into-hibernate-mode

Nonetheless, we should throw out a bad connection and force a refresh, such that the socket factory recovers gracefully from sleep.

See https://github.com/luryus/cloud-sql-jdbc-hang for a reproduction of this issue.

Related to https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/1788.

Whatever we do here, we should port to the AlloyDB Java Connector as well.

enocom:
cc @luryus

This is a problem across Connectors and we do want to improve support for sleep.

luryus:
Hi! I noticed that this (or a very similar issue) got fixed in the Go connector: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/686 

Any chance this could now get ported to this Java library as well?

annafang-google:
Ack. We'll port the fix over shortly.

enocom:
I think this change should match the Go connector -- checking for an invalid cert and blocking until one is available if the cert is in fact expired.

luryus:
Hi @ttosta-google , 

it seems that this is still *not* fixed in v1.16.0 and with #1771.

Even though the library now immediately starts a refresh when it notices that the current cert is expired, the force refresh does not succeed. Logs indicate that it gets stuck in the refresh rate limiter somehow. The library logs "Acquiring rate limiter permit.", but not "Rate limiter permit acquired". All subsequent connection retries then just get the good old `bad_certificate` errors.

This reproduces with both my [reproducer project](https://github.com/luryus/cloud-sql-jdbc-hang) and DBeaver.

Here's sample logs:
```
Feb 17, 2024 1:33:22 PM com.google.cloud.sql.core.Refresher handleRefreshResult
FINE: [<instance name>] Refresh Operation: Completed refresh with new certificate expiration at 2024-02-17T12:33:21Z.
Feb 17, 2024 1:33:22 PM com.google.cloud.sql.core.Refresher handleRefreshResult
[...]

PR #1771, opened 2024-01-10
fix: ensure cert refresh recovers from sleep

ttosta-google:
When a computer wakes from sleep after more than ~1 hour and a client tries to connect, the internal certificate cache will have an expired certificate. As a result, clients who tried to connect after sleep would see bad certificate errors which would resolve only once the remaining refresh duration expired (sleep time + remaining duration).

This commit adds code to check if the client certificate is expired and immediately triggers a blocking refresh if it is. This means that in the case where a cached certificate has expired, it will be discarded and a new one will be retrieved before continuing with the connection.

Fixes https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/1558

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/Connector.java
+++ b/core/src/main/java/com/google/cloud/sql/core/Connector.java
@@ -128,7 +128,17 @@ class Connector {
   }
 
   DefaultConnectionInfoCache getConnection(ConnectionConfig config) {
-    return instances.computeIfAbsent(config, k -> createConnectionInfo(config));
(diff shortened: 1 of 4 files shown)


enocom (changes requested):
This looks great. Let's add a unit test to make sure we don't remove this code in the future.

ttosta-google:
Added a unit test.

enocom (approved):
Thanks for adding a unit test. LGTM.

PR #1779, opened 2024-01-12
chore: correct misconfigured integration test

### LINKED ISSUE #1778: Integration test incorrect. Add test coverage for IAM Auth + Account Impersonation.
### Bug Description

The test `com/google/cloud/sql/mysql/JdbcMysqlJ8ServiceAccountImpersonationIntegrationTests.java` uses the JDBC property `targetPrincipal` instead of `cloudSqlTargetPrincipal`, which means that it is not actually testing service account impersonation. Instead it's using the application default credentials. 

Additionally, it doesn't use IAM Authentication either. So we don't have test coverage for IAM Auth + Account Impersonation. 

### Example code (or command)

_No response_

### Stacktrace

_No response_

### Steps to reproduce?

1. ?
2. ?
3. ?
...

### Environment

1. OS type and version:
2. Java SDK version:
3. Cloud SQL Java Socket Factory version:

### Additional Details

_No response_

hessjcg:
Fixes #1778


enocom:
@hessjcg Let's make sure to mark test-only changes as chores. There's no bug fix here otherwise.

enocom (approved):

PR #1802, opened 2024-01-23
feat: Background refresh stops on unrecoverable Admin API response

### LINKED ISSUE #1797: Failed connection info calls should stop the background refresh
### Feature Description

When an Admin API call fails for an instance, the connectors stop the background refresh in case the failure was caused by a terminal error (instance was deleted, connection string is completely wrong, etc).

ttosta-google:
Fixes #1797

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
@@ -31,6 +31,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningScheduledExecutorService;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
 import java.security.KeyPair;
@@ -44,9 +45,13 @@ import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManagerFactory;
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -59,6 +64,7 @@ class DefaultConnectionInfoRepository implements ConnectionInfoRepository {
   private static final Logger logger =
       LoggerFactory.getLogger(DefaultConnectionInfoRepository.class);
   private final SQLAdmin apiClient;
+  private static final List<Integer> TERMINAL_STATUS_CODE = Arrays.asList(400, 401, 403, 404);
```
enocom, line 67: nit: TERMINAL_STATUS_CODES

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
```diff
@@ -242,13 +244,45 @@ public void create_notAuthorized() throws IOException {
               .withCloudSqlInstance("myProject:myRegion:NotMyInstance")
               .withIpTypes("PRIMARY")
               .build());
-      fail();
+      fail("Expected TerminalException");
+    } catch (TerminalException e) {
```
enocom, line 248: Would `assertThrows` be a good fit here?

core/src/test/java/com/google/cloud/sql/core/RefresherTest.java
```diff
@@ -487,6 +487,65 @@ public void testCloudSqlRefreshesTokenIfExpired() throws Exception {
...
+                  throw new TerminalException("Not authorized");
+              }
+              // refresh 2 and on should return data immediately
+              refreshCount.incrementAndGet();
+              return Futures.immediateFuture(refreshResult);
+            },
+            rateLimiter);
+
+    assertThrows(TerminalException.class, () -> r.getConnectionInfo(TEST_TIMEOUT_MS));
+    assertThat(refreshCount.get()).isEqualTo(1);
```
enocom, line 515: Does this assertion fail before the changes above? Is there a race condition here where this might pass before the refresher's executor would run the next task?

enocom: A comment explaining why this assertion works with the time ut above might help.

enocom (approved):
Minor nit about test assertion. But otherwise LGTM!

core/src/test/java/com/google/cloud/sql/core/InternalConnectorRegistryTest.java
ttosta-google: Updated. Thanks for the suggestion.

core/src/test/java/com/google/cloud/sql/core/RefresherTest.java
ttosta-google: The assertion on line 514 fails before implementing the changes above because no exception is raised by `getConnectionInfo` when the refresher's executor runs the next task.

Added a comment.

PR #1815, opened 2024-01-30
chore: Define dependencyManagement at top-level to centralize the dependency versions used by the submodules

### LINKED ISSUE #818: The top-level pom.xml should define dependencyManagement and subprojects should use it
## Feature Description

The top-level pom.xml should declare dependencyManagement and the subprojects should use the versions defined there.

## Alternatives Considered

N/A

## Additional Context

Without a single place for dependencyManagement, I would have to write duplicate configuration for a dependency in many places. Example: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/805/commits/51f5612252700ca0307fdf58e7d6883d08d8590c

ttosta-google:
Fixes #818


suztomo:
Thank you!

suztomo (approved):

enocom (approved):
nit: Adding some text to the commit message explaining why we're doing this would be nice. Reading between the lines, I think the idea is: dependency management is cheaper at the parent level and not worth breaking up for children pom files.

PR #1861, opened 2024-02-22
fix: Guava dependency missing from jar-with-dependencies

### LINKED ISSUE #1858: Fat "jar-with-dependencies" PostgreSQL JDBC driver is missing Guava dependency inside
### Bug Description

I've built the JDBC postgres socket factory with `mvn -P jar-with-dependencies clean package -DskipTest` as pre README. 

When trying to use the `postgres-socket-factory-1.16.1-SNAPSHOT-jar-with-dependencies.jar`, I get this runtime failure:
```
ERROR: java.lang.NoClassDefFoundError: com/google/common/util/concurrent/internal/InternalFutureFailureAccess
```

When I remove `<scope>test</scope>` from Guava entry in `jdbc/postgres/pom.xml`, the resulting JAR file works. I'm however not sure if that is the right fix or if the Guava omission is not intentional, so I didn't create a PR.

### Example code (or command)

_No response_

### Stacktrace

_No response_

### Steps to reproduce?

1. Build with `mvn -P jar-with-dependencies clean package -DskipTest`
2. Use in some project with JDBC driver that is not already using Guava, eg. Keycloak
3. Runtime error with not found class

### Environment

1. OS type and version: Linux
2. Java SDK version: 17
3. Cloud SQL Java Socket Factory version: 1.16.1

### Additional Details

_No response_

Tasssadar:
Doing the solution suggested in  #1858, but for all the jdbc drivers, as they all seem to have the same problem.

It compiles correctly, test also compile but I'm unable to run them locally.

Fixes #1858

### DIFF
--- a/jdbc/mariadb/pom.xml
+++ b/jdbc/mariadb/pom.xml
@@ -58,11 +58,6 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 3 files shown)


ttosta-google (approved):
LGTM

PR #1862, opened 2024-02-22
chore: Add integration tests for Quota Project feature.

ttosta-google:
Add new integration tests for Quota Project feature:

1. Quota Project + IAM Authentication
2. Quota Project + Service Account Impersonation

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -101,6 +101,7 @@ jobs:
           SQLSERVER_PASS:${{ secrets.GOOGLE_CLOUD_PROJECT }}/SQLSERVER_PASS
           SQLSERVER_DB:${{ secrets.GOOGLE_CLOUD_PROJECT }}/SQLSERVER_DB
           IMPERSONATED_USER:${{ secrets.GOOGLE_CLOUD_PROJECT }}/IMPERSONATED_USER
+          QUOTA_PROJECT:${{ secrets.GOOGLE_CLOUD_PROJECT }}/QUOTA_PROJECT
     - name: Run tests
       env:
         MYSQL_CONNECTION_NAME: '${{ steps.secrets.outputs.MYSQL_CONNECTION_NAME }}'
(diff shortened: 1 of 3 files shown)


enocom (commented):
How would these tests fail? For instance, do we have unit level coverage that we're wiring up the quota project correctly?

Is there any observable effect that setting the quota project has? For example, could we query the quota project for consumption?

ttosta-google:
These tests would fail to fetch the certificate.

`POST https://sqladmin.googleapis.com/sql/v1beta4/projects/***instances/***:generateEphemeralCert`

Some failures:

1. Invalid quota project

```
Error code: 400 Bad Request
Reason: USER_PROJECT_DENIED
Message: Project 'projects/QUOTA-PROJECT' not found or deleted."
```

2. Cloud Admin API is disabled in the quota project

```
Error code: 403 Forbidden
Reason: SERVICE_DISABLED
Message: Cloud SQL Admin API has not been used in project <QUOTA-PROJECT> before or it is disabled.
```

3. Service account does not have the required permission

```
Error code: 403 Forbidden 
Reason: USER_PROJECT_DENIED
Message: Caller does not have required permission to use project <QUOTA-PROJECT>. Grant the caller the roles/serviceusage.serviceUsageConsumer role, or a custom role with the serviceusage.services.use permission.
```

After running the tests, I manually checked the Cloud SQL Admin API metrics in the quota project https://pantheon.corp.google.com/apis/api/sqladmin.googleapis.com/metrics?project=QUOTA-PROJECT and saw the numbers going up. 

Checking the Admin API metrics in the test is not accurate as the e2e tests for different envs (OS x Java version) are executed concurrently.

Please let me know your thoughts.

enocom (approved):
Thanks for the clarification.

jackwotherspoon:
@ttosta-google adding to what Eno previously said:

> Is there any observable effect that setting the quota project has? For example, could we query the quota project for consumption?

Is there a way within the test to verify the quota project that being set is actually being used? Otherwise I fear these tests could regress in the future...

Also, I don't think our other repos have this test, @enocom @ttosta-google should we file a bug to track adding a similar test to other repos?

ttosta-google:
PR https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/1866: added another integration test to verify that the quota project is used when sending requests to Cloud SQL Admin API.

It is not possible to verify it on the integration tests added under this PR because they only have access to the Spring Boot Connection pool.

Agree on adding similar tests to other repos.

PR #1943, opened 2024-04-15
chore: fix JDBC readme SQL Server dependency

jackwotherspoon:
JDBC README shows incorrect dependency for SQL Server usage.

I believe it should be `cloud-sql-connector-jdbc-sqlserver`?


ttosta-google (approved):
Good catch. Thanks!

Issue #1940, opened 2024-04-12
support jdbc connection with unix domain socket in native image


czp3009:
### Feature Description

I'm sorry I don't really know where this issue should be submit, so I'll submit it here first, and let me know if it shouldn't be.

I've looked at the [previous archived repositories](https://github.com/GoogleCloudPlatform/native-image-support-java), but I don't know what repositories the previous functionality has been moved to.

I've submitted a PR to that repository before: https://github.com/GoogleCloudPlatform/native-image-support-java/pull/304

That PR add unix domain socket supporting in graalvm native images.

But now when I add dependency according to the new documentation, I still get the same error as before this PR

```
java.lang.ClassNotFoundException: jnr.ffi.provider.jffi.Provider
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.hub.ClassForNameSupport.forName(ClassForNameSupport.java:123) ~[na:na]
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.hub.ClassForNameSupport.forName(ClassForNameSupport.java:87) ~[na:na]
        at java.base@17.0.10/java.lang.Class.forName(DynamicHub.java:1322) ~[report-service:na]
        at java.base@17.0.10/java.lang.Class.forName(DynamicHub.java:1285) ~[report-service:na]
        at java.base@17.0.10/java.lang.Class.forName(DynamicHub.java:1278) ~[report-service:na]
        at jnr.ffi.provider.FFIProvider$SystemProviderSingletonHolder.getInstance(FFIProvider.java:68) ~[na:na]
        at jnr.ffi.provider.FFIProvider$SystemProviderSingletonHolder.<clinit>(FFIProvider.java:57) ~[na:na]
        at jnr.ffi.provider.FFIProvider.getSystemProvider(FFIProvider.java:35) ~[report-service:na]
        at jnr.ffi.Runtime$SingletonHolder.<clinit>(Runtime.java:109) ~[na:na]
        at jnr.ffi.Runtime.getSystemRuntime(Runtime.java:70) ~[report-service:na]
        at jnr.unixsocket.SockAddrUnix.<init>(SockAddrUnix.java:46) ~[report-service:na]
        at jnr.unixsocket.SockAddrUnix$DefaultSockAddrUnix.<init>(SockAddrUnix.java:208) ~[na:na]
[...]

PR #1960, opened 2024-04-23
test: Add a unix socket connector test. Part of #1940

hessjcg:
This adds a connector test that exercises the connector's unix socket behavior. This should expose that Graalvm reflection is not configured correctly.

See #1940

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -60,6 +60,12 @@
 
     <dependency>
       <groupId>com.github.jnr</groupId>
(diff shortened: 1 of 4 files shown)


enocom:
I'd suggest running this with https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/1961 at least once here in the PR before merging just in case the build breaks in a surprising way.

hessjcg:
Closing. This change will be combined into #1961.

PR #1966, opened 2024-05-03
chore: Add refreshStrategy connector configuration (part of #992)

hessjcg:
This adds the new JDBC and R2DBC configuration property `refreshStrategy` with valid values `lazy` or `background`. 
For R2DBC, the configuration option is named 'REFRESH_STRATEGY' to match formatting of the 
other configuration options. 

Part of #992.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/RefreshStrategy.java
+++ b/core/src/main/java/com/google/cloud/sql/RefreshStrategy.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/google/cloud/sql/RefreshStrategy.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql;
+
+/** Enum for supported refresh strategies. */
+public enum RefreshStrategy {
+  BACKGROUND,
+  LAZY
```
enocom, line 22: Since Lazy doesn't do anything yet, shall we leave this out and introduce it again when it actually triggers a change?

enocom (approved):

PR #1967, opened 2024-05-03
chore: Rename refresher to RefreshAheadRefresher, part of #992.

hessjcg:
Rename classes to highlight the difference between the different refresh strategies.  

Part of lazy refresh #992.


jackwotherspoon (commented):
is `Refresher` a widely used term or something we made up for the Connectors? I'm not opposed to it but `RefreshAheadRefresher` is quite a mouthful...

enocom (commented):
Maybe just drop the Refresher and call it a "RefreshAhead" -- the word "strategy" is out there too.

hessjcg:
Good point on the names. Maybe we call the interface `RefreshStrategy` and the two classes `LazyStrategy` and `RefreshAheadStrategy`.

jackwotherspoon:
> Good point on the names. Maybe we rename the `Refresher` interface to `RefreshStrategy` and the two implementation classes are named `LazyStrategy` and `RefreshAheadStrategy`.

That works for me

jackwotherspoon (approved):
LGTM

nit: should we use `refactor:` here instead of `chore:`?

PR #1968, opened 2024-05-06
chore: Introduce the RefreshStrategy interface (part of #992)

hessjcg:
Introduces the RefreshStrategy interface and updates RefreshAheadStrategy to implement that interface. 

Part of #992

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
@@ -35,7 +35,7 @@ import javax.net.ssl.SSLSocket;
 class DefaultConnectionInfoCache {
   private final AccessTokenSupplier accessTokenSupplier;
   private final CloudSqlInstanceName instanceName;
-  private final RefreshAheadStrategy refreshStrategy;
+  private final RefreshStrategy refreshStrategy;
   private final ConnectionConfig config;
 
   /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/cloud/sql/core/RefreshStrategy.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.google.cloud.sql.core;
+
+/** Provide the refresh strategy to the DefaultConnectionInfoCache. */
+public interface RefreshStrategy {
+  ConnectionInfo getConnectionInfo(long timeoutMs);
+
+  void forceRefresh();
+
+  void refreshIfExpired();
```
jackwotherspoon, line 25: is `refreshIfExpired` needed on the base interface?

we don't have this on the Go interface i think https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/768?

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoCache.java
```diff
@@ -35,7 +35,7 @@
 class DefaultConnectionInfoCache {
   private final AccessTokenSupplier accessTokenSupplier;
   private final CloudSqlInstanceName instanceName;
-  private final RefreshAheadStrategy refreshStrategy;
+  private final RefreshStrategy refreshStrategy;
```
jackwotherspoon, line 38: should the default not still be `RefreshAheadStrategy`?

hessjcg: Technically yes, it could because there is exactly one implementation. But because this PR is about introducing the RefreshStrategy interface class, I think it makes sense update the code where the interface gets used together.

core/src/main/java/com/google/cloud/sql/core/RefreshStrategy.java
hessjcg: Yes. This is important for some edge cases in the RefreshAheadStrategy. It is a noop in the LazyRefresher.

enocom: We do the equivalent in the dialer in the Go Connector. In theory, we could hide that inside the cache.

jackwotherspoon (approved):
LGTM, let's get @enocom to approve as well prior to merging

PR #1970, opened 2024-05-07
chore: Add LazyRefreshStrategy implementing the lazy refresh strategy. (part of #992)

hessjcg:
This adds code and unit tests implementing the LazyRefreshStrategy.


core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,152 @@
...
+                name));
+        fetchConnectionInfo(timeoutMs);
+      }
+      return connectionInfo;
+    }
+  }
+
+  private void fetchConnectionInfo(long timeoutMs) {
+    synchronized (connectionInfoGuard) {
+      final long finishBeforeTimestamp = System.currentTimeMillis() + timeoutMs;
```
enocom, line 78: Why the loop and rate limiter here? I would expect to see just a simple call and caching of the result. If we want to add retries, we should do it down below in the code that actually calls the API.

hessjcg: It is easier to keep the separation of concerns consistent when the retry and rate limit are in the RefreshStrategy, not the `ConnectionInfoRepository`.

The `ConnectionInfoRepository` responsibility is simply to attempt to get the latest information from the API. If it fails, then it does not attempt to retry. 

The `RefreshStrategy` responsibility is to handle timing and concurrency. This includes the refresh delays, rate limits, and retries, which all interact with each other. By isolating all timing and concurrency control in the RefreshStrategy, we can unit test and verify it. 

The existing `RefreshAheadStrategy` implementation handles refresh delay, rate limit, and retry with Futures. This gets really weird so for legibility, it needs to all be in one place. 

The `LazyRefreshStrategy` implementation is much simpler because it is synchronous, but needs to have the same set of responsibilities: delay, rate limit, retry.

enocom: I could see how retries being here makes sense, but the rate limiter probably doesn't belong here. If you look at the reference implementation in Go, we rate limit in the refresh ahead cache, but not in the lazy cache or the underlying API client that includes retries.

hessjcg: You are correct that this forces us to implement retries (and rate limits) twice. But that is necessary because of this Java language constraint:  In Java, the waiting logic for retry and rate limit logic is implemented differently for Futures in RefreshAheadStrategy than for synchronous calls in LazyRefreshStrategy. 

The Futures implementation releases the limited thread pool threads while it waits for a retry or to acquire a rate limiter lease.  Because the Go implementation uses goroutines, the go connector can use the same implementation using blocking calls for both the Refresh Ahead and the Lazy Refresh strategies.

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,152 @@
...
+import org.slf4j.LoggerFactory;
+
+/** RefreshStrategy that implements the lazy refresh strategy. */
+public class LazyRefreshStrategy implements RefreshStrategy {
+  private final Logger logger = LoggerFactory.getLogger(LazyRefreshStrategy.class);
+
+  private final String name;
+  private final Supplier<ConnectionInfo> refreshOperation;
+
+  private final RateLimiter rateLimiter;
```
jackwotherspoon, line 33: Eno mentioned this in a below comment, but does the Lazy Refresh need a rate limiter? I would think no. Since the rate limiter is meant to slow background refresh calls I think we don't need it here.

We should not rate limit for lazy refresh as it may provide unintended behaviour.

jackwotherspoon (changes requested):

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+      }
+      return connectionInfo;
+    }
+  }
+
+  private void fetchConnectionInfo() {
+    synchronized (connectionInfoGuard) {
+      logger.debug(String.format("[%s] Refresh Operation: Starting refresh operation.", name));
+      try {
+        this.connectionInfo = this.refreshOperation.get();
```
enocom, line 75: Does the supplier time out on a failed get?

enocom (approved):

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
hessjcg: Yes. The api client has a default timeout.

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
hessjcg: RateLimiter has been removed.

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+        throw new IllegalStateException("Named connection closed");
+      }
+
+      if (connectionInfo == null) {
+        logger.debug(
+            String.format(
+                "[%s] Client certificate has expired. Starting next refresh operation immediately.",
+                name));
+        fetchConnectionInfo();
+      }
```
jackwotherspoon, line 59: when does this case occur?

jackwotherspoon: looks like it occurs when we invalidate the cache, can we have a better debug message?

jackwotherspoon: maybe "Cache is empty. Starting next refresh immediately."

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+  /** Force a new refresh of the instance data if the client certificate has expired. */
+  @Override
+  public void forceRefresh() {
+    // invalidate connectionInfo so that the next call to getConectionInfo() will retrieve
+    // fetch new data.
+    synchronized (connectionInfoGuard) {
+      if (closed) {
+        throw new IllegalStateException(String.format("[%s] Named connection closed.", name));
+      }
+      this.connectionInfo = null;
```
jackwotherspoon, line 99: should we debug log here?

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+      if (closed) {
+        throw new IllegalStateException("Named connection closed");
+      }
+    }
+  }
+
+  @Override
+  public void close() {
+    synchronized (connectionInfoGuard) {
+      closed = true;
```
jackwotherspoon, line 116: worth a debug log?

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+              "[%s] Refresh Operation: Completed refresh with new certificate expiration at %s.",
+              name, connectionInfo.getExpiration().toString()));
+    }
+  }
+
+  /** Force a new refresh of the instance data if the client certificate has expired. */
+  @Override
+  public void forceRefresh() {
+    // invalidate connectionInfo so that the next call to getConectionInfo() will retrieve
+    // fetch new data.
```
jackwotherspoon, line 94: nit: choose one of retrieve/fetch they are synonyms
```suggestion
    // invalidate connectionInfo so that the next call to getConectionInfo() will retrieve
    // fetch new data.
```

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
hessjcg: Updated

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+                name));
+        fetchConnectionInfo();
+      }
+      return connectionInfo;
+    }
+  }
+
+  private void fetchConnectionInfo() {
+    synchronized (connectionInfoGuard) {
+      logger.debug(String.format("[%s] Refresh Operation: Starting refresh operation.", name));
```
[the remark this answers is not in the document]
hessjcg: Good idea!  Updated.

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/LazyRefreshStrategy.java
hessjcg: Fixed.

jackwotherspoon (approved):
LGTM

PR #1996, opened 2024-05-22
feat: Configure java connector to check CN instance name. Fixes #1995

### LINKED ISSUE #1995: Check that the server certificate CN matches the instance name
### Feature Description

During the TLS authentication, the server certificate CN field should match the instance name. Add a custom TrustManager that will check the CN field and reject it if the name does not match the instance name.

hessjcg:
This code will make that the connector checks that the server certificate's subject CN field matches the instance name
before allowing the TLS handshake to proceed.

This introduces new classes that are always installed as trust managers on all TLS connections to ensure that the TLS logic correctly checks that the server certificate's CN field matches the instance.

-  `InstanceCheckingTrustManager` - Delegates to default TLS trust manager, then checks the CN field.
-  `InstanceCheckingTrustManagerFactory` and `InstanceCheckingTrustManagerFactorySpi` - Installs
     our custom trust manager into connector TLS sockets.

The `ConscryptWorkaroundTrustManagerFactory` and `ConscryptWorkaroundTrustManagerFactorySpi` classes are no longer necessary. Logic to detect Conscrypt and use the workaround `ConscryptWorkaroundTrustManager` moves into `InstanceCheckingTrustManagerFactorySpi`.


enocom (commented):
Do we have a test that verifies a common name mismatch fails the TLS handshake?

hessjcg:
I added a test for a CN mismatch.

jackwotherspoon:
> @jackwotherspoon We need to manually test this with a domain-scoped instance. Can you do that?

@jonath sure, for manually testing should I just run the integration tests? Will that suffice?

hessjcg:
@jackwotherspoon: Yes, it will be sufficient to run the integration tests at the domain-scoped instance.

core/src/test/java/com/google/cloud/sql/core/CloudSqlCoreTestingBase.java
```diff
@@ -62,6 +62,8 @@ public class CloudSqlCoreTestingBase {
 
   ListenableFuture<KeyPair> clientKeyPair;
 
+  public CloudSqlCoreTestingBase() {}
```
enocom, line 65: FWIW In my opinion, this base class deserves to be split up into small composable units. As is, it's an awkward tool.

core/src/test/java/com/google/cloud/sql/core/ConnectorTest.java
```diff
@@ -80,6 +81,32 @@ public void create_throwsErrorForInvalidInstanceName() throws IOException {
...
   }
 
+  @Test
+  public void create_throwsErrorForInvalidTlsCommonNameMismatch()
+      throws IOException, InterruptedException {
+    // The server TLS certificate matches myProject:myRegion:myInstance
+    FakeSslServer sslServer = new FakeSslServer();
+    ConnectionConfig config =
+        new ConnectionConfig.Builder()
+            .withCloudSqlInstance("myProject:myRegion:otherInstance")
```
enocom, line 91: nit: let's make this blatantly wrong, e.g., "wrongwrongwrong" or something similar that catches the eye.

enocom (approved):

core/src/test/java/com/google/cloud/sql/core/CloudSqlCoreTestingBase.java
hessjcg: +1. Maybe in a different PR.

core/src/test/java/com/google/cloud/sql/core/ConnectorTest.java
hessjcg: Fixed.

PR #1997, opened 2024-05-23
refactor: Refactor RefreshAheadConnectionInfoCache. Part of #992.

hessjcg:
This makes a number of refactoring changes to align the Java connector with other implementations. 
- Introduce a ConnectionInfoCache interface 
- Rename DefaultConnectionInfoCache to RefreshAheadConnectionInfoCache
- Update and simplify instantiation logic of RefreshAheadConnectionInfoCache


core/src/main/java/com/google/cloud/sql/core/RefreshAheadConnectionInfoCache.java
```diff
@@ -0,0 +1,70 @@
...
+   * Initializes a new Cloud SQL instance based on the given connection name. + * Initializes a new
+   * Cloud SQL instance based on the given connection name using the background + * refresh
+   * strategy.
+   *
+   * @param config instance connection name in the format "PROJECT_ID:REGION_ID:INSTANCE_ID"
+   * @param connectionInfoRepository Service class for interacting with the Cloud SQL Admin API
+   * @param executor executor used to schedule asynchronous tasks
+   * @param keyPair public/private key pair used to authenticate connections
+   */
+  public static RefreshAheadConnectionInfoCache newInstance(
```
hessjcg, line 40: I have to use a static factory method instead of a constructor because it has to instantiate an AccessTokenSupplier and RefreshAheadStrategy before it calls the constructor.,

enocom (commented):
What if the ConnectinInfoCache just returned the connection info instead of creating SSL sockets? Recently, in Go and Python we've been pulling apart the thing that caches connection info (e.g., the "cache") from the thing that creates SSL sockets (e.g., the "connector"). That's what we're doing in AlloyDB and it's working well: https://github.com/GoogleCloudPlatform/alloydb-java-connector/blob/main/alloydb-jdbc-connector/src/main/java/com/google/cloud/alloydb/ConnectionInfoCache.java

hessjcg:
I'll remove the `createSslSocket()` method from the ConnectionInfoCache. It really is only a convenience method anyway.

core/src/main/java/com/google/cloud/sql/core/BaseConnectionInfoCache.java
```diff
@@ -19,62 +19,39 @@
...
-      ConnectionConfig config,
-      ConnectionInfoRepository connectionInfoRepository,
-      CredentialFactory tokenSourceFactory,
-      ListeningScheduledExecutorService executor,
-      ListenableFuture<KeyPair> keyPair,
-      long minRefreshDelayMs) {
-    this.instanceName = new CloudSqlInstanceName(config.getCloudSqlInstance());
-    this.config = config;
-
+abstract class BaseConnectionInfoCache implements ConnectionInfoCache {
```
enocom, line 30: The base class is delegating all calls to the refresh strategy. What if we avoid introducing more indirection and just had the implementations of ConnectionInfoCache call through to their refresher?

core/src/main/java/com/google/cloud/sql/core/ConnectionInfoCache.java
```diff
@@ -0,0 +1,40 @@
...
+   */
+  ConnectionMetadata getConnectionMetadata(long timeoutMs);
+
+  void forceRefresh();
+
+  void refreshIfExpired();
+
+  RefreshStrategy getRefresher();
+
+  CloudSqlInstanceName getInstanceName();
```
enocom, line 37: This is used only in test. Let's remove the method here and either drop the usages in test or find a more direct way at getting at the information the test wants.

core/src/main/java/com/google/cloud/sql/core/ConnectionInfoCache.java
```diff
@@ -0,0 +1,40 @@
...
+   * @throws IllegalArgumentException If the instance has no IP addresses matching the provided
+   *     preferences.
+   */
+  ConnectionMetadata getConnectionMetadata(long timeoutMs);
+
+  void forceRefresh();
+
+  void refreshIfExpired();
+
+  RefreshStrategy getRefresher();
```
enocom, line 35: This is a test-only method. How can we avoid adding it here?

core/src/main/java/com/google/cloud/sql/core/RefreshAheadConnectionInfoCache.java
```diff
@@ -0,0 +1,70 @@
...
+      ConnectionConfig config,
+      ConnectionInfoRepository connectionInfoRepository,
+      CredentialFactory tokenSourceFactory,
+      ListeningScheduledExecutorService executor,
+      ListenableFuture<KeyPair> keyPair,
+      long minRefreshDelayMs) {
+
+    AccessTokenSupplier accessTokenSupplier =
+        BaseConnectionInfoCache.newAccessTokenSupplier(config, tokenSourceFactory);
```
enocom, line 49: This static method is making our lives harder. Why don't we pass in the access token supplier from where we instantiate the cache?

enocom (changes requested):

core/src/main/java/com/google/cloud/sql/core/ConnectionInfoCache.java
hessjcg: I've removed it from the interface.

core/src/main/java/com/google/cloud/sql/core/BaseConnectionInfoCache.java
hessjcg: I removed the base class.

core/src/main/java/com/google/cloud/sql/core/ConnectionInfoCache.java
hessjcg: I removed this method too.

core/src/main/java/com/google/cloud/sql/core/RefreshAheadConnectionInfoCache.java
hessjcg: Refactored, removing the base class, so that this can be a normal constructor.

core/src/main/java/com/google/cloud/sql/core/ConnectionInfo.java
```diff
@@ -51,4 +52,29 @@ Map<IpType, String> getIpAddrs() {
   SslData getSslData() {
     return sslData;
   }
+
+  ConnectionMetadata toConnectionMetadata(
```
hessjcg, line 56: Moved here because `BaseConnectionInfoCache` was deleted. 

(Additional context: `ConnectionMetadata` is part of the legacy public Java API contract.)

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
```diff
@@ -43,6 +44,14 @@ class DefaultAccessTokenSupplier implements AccessTokenSupplier {
   private final int retryCount;
   private final Duration retryDuration;
 
+  static AccessTokenSupplier newInstance(AuthType authType, CredentialFactory tokenSourceFactory) {
```
hessjcg, line 47: Moved here because BaseConnectionInfoCache was deleted.

core/src/main/java/com/google/cloud/sql/core/ConnectionInfo.java
enocom: This seems like the perfect place for such concerns -- basically the connection info data class.

core/src/main/java/com/google/cloud/sql/core/ConnectionInfoCache.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+/** ConnectionInfoCache is the contract for a caching strategy for ConnectionInfo. */
+public interface ConnectionInfoCache {
```
enocom, line 20: Do we want this to be public?

core/src/main/java/com/google/cloud/sql/core/DefaultAccessTokenSupplier.java
enocom: Perfect spot.

enocom (approved):
There's one use of public to probably remove. Otherwise LGTM.

core/src/main/java/com/google/cloud/sql/core/ConnectionInfoCache.java
hessjcg: I removed `public`.

PR #2002, opened 2024-05-28
test: Script creation of TLS keys and certificates used in tests. Part of #1995

hessjcg:
Update our tests so that the keys and certificates are generated at test time. This will make testing more flexible
and easier to update.


enocom (commented):
Cf. https://github.com/GoogleCloudPlatform/alloydb-java-connector/blob/main/alloydb-jdbc-connector/src/test/java/com/google/cloud/alloydb/TestCertificates.java where we generate certs on a per test-run basis. This removes any need to regenerate certs when they expire.

hessjcg:
That could be a good next step to implement when build customer-defined SAN records.  For now, I just want a basic replacement of what exists so that I can implement the CN check.

enocom (commented):
I know what a pain setting up openssl calls can be. This PR seems like more of a lateral change rather than an improvement. What would it take it to translate the openssl calls into Java? If we did that, it would make maintenance of these certs much easier.

hessjcg:
I rewrote it following the model from AlloyDB.

enocom (approved):
Thank you for doing this. It's going to be a big lift as time passes.

PR #2041, opened 2024-07-15
feat: Retry API calls that return a 500 error

### LINKED ISSUE #2029: Add retries around 50X errors
Port https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/781 to Java Connector.

We have seen the connect API occasionally give 502 errors, we should be retrying 50X errors to make the Java Connector more resilient.

hessjcg:
This updates the retry logic for API calls made by the connector: 

- Attempts to get an authentication token from the Google auth API are always retried 5 times.
- SQL Admin API requests are retried up to 5 times if the http response status code is a 5xx. Otherwise they are fatal on the first error.

This also implements the exponential backoff logic defined used in the go connector.
See: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/781

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/ApiClientRetryingCallable.java
+++ b/core/src/main/java/com/google/cloud/sql/core/ApiClientRetryingCallable.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import com.google.api.client.http.HttpResponseException;
+import java.util.concurrent.Callable;
+
+/**
+ * Extends RetryingCallable with logic to only retry on HTTP errors with error codes in the 5xx
+ * range.
+ *
+ * @param <T> the return value for Callable
+ */
+class ApiClientRetryingCallable<T> extends RetryingCallable<T> {
+
+  /**
+   * Construct a new RetryLogic.
+   *
+   * @param callable the callable that should be retried
+   */
+  public ApiClientRetryingCallable(Callable<T> callable) {
+    super(callable);
+  }
+
+  /**
+   * Returns false indicating that there should be another attempt if the exception is an HTTP
+   * response with an error code in the 5xx range.
+   *
+   * @param e the exception
+   * @return false if this is a http response with a 5xx status code, otherwise true.
+   */
+  @Override
+  protected boolean isFatalException(Exception e) {
+    // Only retry if the error is an HTTP response with a 5xx error code.
+    if (e instanceof HttpResponseException) {
+      HttpResponseException re = (HttpResponseException) e;
+      return re.getStatusCode() < 500;
+    }
+    // Otherwise this is a fatal exception, no more tries.
+    return true;
+  }
+}
--- a/core/src/main/java/com/google/cloud/sql/core/RetryingCallable.java
+++ b/core/src/main/java/com/google/cloud/sql/core/RetryingCallable.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.sql.core;
 
-import java.time.Duration;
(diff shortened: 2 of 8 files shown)


core/src/main/java/com/google/cloud/sql/core/ApiClientRetryingCallable.java
```diff
@@ -0,0 +1,55 @@
...
+  public ApiClientRetryingCallable(Callable<T> callable) {
+    super(callable);
+  }
+
+  /**
+   * Returns false indicating that there should be another attempt if the exception is an HTTP
+   * response with an error code in the 500 range.
+   *
+   * @param e the exception
+   * @return false if this is a http response with a 500 status code, otherwise true.
```
jackwotherspoon, line 44:
```suggestion
   * response with an error code in the 5xx range.
   *
   * @param e the exception
   * @return false if this is a http response with a 5xx status code, otherwise true.
```

core/src/main/java/com/google/cloud/sql/core/RetryingCallable.java
```diff
@@ -25,59 +24,57 @@
  * The sleep duration is chosen randomly in the range [sleepDuration, sleepDuration * 2] to avoid
  * causing a thundering herd of requests on failure.
  *
+ * <p>exponentialBackoff calculates a duration based on the attempt i.
+ *
+ * <p>The formula is: base * multi^(attempt + 1 + random)
+ *
+ * <p>With base = 200ms and multi = 1.1618, and random = [0.0, 1.0), the backoff values would fall
```
jackwotherspoon, line 31: nit: value of multi should be 1.618
```suggestion
 * <p>With base = 200ms and multi = 1.618, and random = [0.0, 1.0), the backoff values would fall
```

core/src/test/java/com/google/cloud/sql/core/ApiClientRetryingCallableTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class ApiClientRetryingCallableTest {
+  @Test
+  public void testApiClientRetriesOn500Error() throws Exception {
+    AtomicInteger counter = new AtomicInteger(0);
+    ApiClientRetryingCallable<Integer> c =
+        new ApiClientRetryingCallable<>(
+            () -> {
+              int attempt = counter.incrementAndGet();
+              if (attempt < 3) {
```
jackwotherspoon, line 35: any reason for choosing 3 instead of 5 here?

jackwotherspoon (approved):
couple nits, will let @enocom review for Java readability

should we make this a PR a `fix:`?

enocom:
This should be a "feature" -- it will have a noticeable usability change.

core/src/main/java/com/google/cloud/sql/core/ApiClientRetryingCallable.java
```diff
@@ -0,0 +1,55 @@
...
+   */
+  @Override
+  protected boolean isFatalException(Exception e) {
+    // Only retry if the error is an HTTP response with a 5xx error code.
+    if (e instanceof HttpResponseException) {
+      HttpResponseException re = (HttpResponseException) e;
+      return re.getStatusCode() < 500 || re.getStatusCode() > 599;
```
enocom, line 51: When would ` re.getStatusCode() > 599` apply? I think  `re.getStatusCode() < 500` will get the job done.

core/src/main/java/com/google/cloud/sql/core/ApiClientRetryingCallable.java
hessjcg: Fixed.

hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/ApiClientRetryingCallable.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/RetryingCallable.java
hessjcg: Fixed.

core/src/test/java/com/google/cloud/sql/core/ApiClientRetryingCallableTest.java
hessjcg: I wanted to demonstrate that it returns after a successful attempt before exhausting the total number of retries. 

I added another test that always fails to show that it stops at 5 attempts and then fails.

hessjcg (commented):
I also added a new test to `ConnectorTest`. This test the SqlAdmin client library actually responds correctly and the ApiClientRetryingCallable works as expected when the service intermittently responds with 502 errors.

Issue #2043, opened 2024-07-18
Automatically configure and fail-over connections using DNS


hessjcg:
### Feature Description

The java connector may be configured to connect to a Cloud SQL instance using a DNS name.

When the JDBC connection properties does not contain the `cloudSqlInstance` property,  the connector will attempt to use the hostname from the JDBC URL to resolve name using DNS. The connector will query the local DNS for a SRV record. The SRV should contain an instance name matching that DNS name. The connector will use the instance name from the SRV record to connect to the Cloud SQL instance.

The connector will periodically query that DNS. If the instance name in the DNS record changes, the connector will terminate all open connections to the old instance. Then, it will reconfigure itself to open connections to the new instance.

This enables several DNS-related features by allowing configuration through DNS.

Users can configure their database clients using the domain name instead of the instance name.
Users can fail-over database clients when they change the DNS entry for the database instance.

This is the Java implementation of https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/842

derek-gfs:
Does this prevent split brain?

jackwotherspoon:
> Does this prevent split brain?

@derek-gfs  Thanks for the question! 😄 

@hessjcg who is driving DR support here is OOO until next week, but will respond here when he is back.

derek-gfs:
Hi All - Any updates on the implementation of this?

Since write endpoints are now in preview, I am eager to give this a try.

jackwotherspoon:
> Hi All - Any updates on the implementation of this?
> 
> Since write endpoints are now in preview, I am eager to give this a try.

@derek-gfs Thanks for showing your interest here! We really appreciate it 😄 

I believe @hessjcg will be picking back up this work in the coming week or two and getting these PRs across the line.

We are still waiting on a backend change to fully roll out before the full support of the feature in the Connectors will be possible.

Issue #1153, opened 2023-02-13
Support DNS names for `cloudSqlInstance`


RB14:
## Feature Description
Hello, 

We have a use case in our application where our database instance is being replaced every day and we need our application to reconnect to it automatically when that happens. The approach we took for the solution is that we created a DNS record which points at the **private ip** of our Cloud SQL instance, and configured the connection pool of our app to use that DNS name. When the instance is being replaced, the process that replaced it would also update the DNS record to point to the private IP of the new instance.

The problem with this approach is that we can't use Cloud SQL Proxy (or the various connectors) for the connection, since Cloud SQL Proxy requires to specify the instance **name** upon **initialization** of the connector, or, alternatively, requires to **restart** the `cloud_sql_proxy` process, which is external to the the application, may require a short downtime, and makes it  more difficult.

I would like to propose a feature for the Cloud SQL Connector library, which would allow to specify the `cloudSqlInstance` via a DNS name. That DNS would be of type [`TXT`](https://en.wikipedia.org/wiki/TXT_record), and its content will be the same content we currently have to pass to `cloudSqlInstance`. The way the connector would work is that every time it is asked for a new connection, it will **first** resolve the DNS name, extract the `TXT` record from it, and then pass it to the [`CoreFactory#connect`](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/jdbc/postgres/src/main/java/com/google/cloud/sql/postgres/SocketFactory.java#L77) method (just like today). I'm not sure if that DNS name should be passed in the conventional `host` parameter, or in another custom parameter, or maybe even to the same `cloudSqlInstance` parameter where you currently expect the instance name, but I'm leaving it for you to decide.

## Alternatives Considered
We have considered several alternatives, but all of them require changes to our architecture (which currently relies on the DNS name)... including restarting the connection pool from the outside, or restarting the `cloud_sql_proxy`... all these solutions are more complicated for us, and we would prefer it we can stick to the DNS solution.

## Additional Context
Using [dynamic] DNS records to point at network resources provides several benefits, but the most significant one is the ability to abstract away the resource from its clients. It allows us to make changes to the resource (e.g. replacing the db instance), while keeping the actual change transparent from the user (just change the DNS name to point at the newly located resource). All that assuming the changes we do are backward compatible on the API level.

RB14:
[...]

PR #2044, opened 2024-07-18
feat: Automatically configure connections using DNS. Part of #2043.

hessjcg:
A connection may be configured to using a DNS name instead of an instance name. If the cloudSqlInstance property is
set to be a domain name instead of an instance name, the connector will look up a SRV record for that name and
connect to that instance.

There should be exactly 1 SRV for the database instance. The connector will always use the SRV record with the
highest priority.

Part of #2043

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/InstanceConnectionNameResolver.java
+++ b/core/src/main/java/com/google/cloud/sql/core/InstanceConnectionNameResolver.java
@@ -16,10 +16,15 @@
 
 package com.google.cloud.sql.core;
 
-import java.util.Collection;
-import javax.naming.NameNotFoundException;
+/** Resolves the Cloud SQL Instance from the configuration name. */
+interface InstanceConnectionNameResolver {
 
-/** Wraps the Java DNS API. */
-interface DnsResolver {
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/google/cloud/sql/core/InstanceConnectionNameResolver.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+/** Resolves the Cloud SQL Instance from the configuration name. */
+interface InstanceConnectionNameResolver {
```
enocom, line 20: I think it's fine to make this package private for now -- but you'll want to plan for making this interface public so that callers can configure their own resolver.

enocom: See https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/1153 as an example of someone who already asked for this.

PR #2045, opened 2024-07-18
chore: Create a new DnsResolver class that wraps the Java DNS API

hessjcg:
This is a prerequisite for other implementation steps in the Automatic Failover with DNS Configuration story.

Part of #2043.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/DnsResolver.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DnsResolver.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import java.util.Collection;
+import javax.naming.NameNotFoundException;
+
+interface DnsResolver {
+  Collection<String> resolveTxt(String domainName) throws NameNotFoundException;
+}
--- a/core/src/main/java/com/google/cloud/sql/core/JndiDnsResolver.java
+++ b/core/src/main/java/com/google/cloud/sql/core/JndiDnsResolver.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


enocom (commented):
LGTM.

I'd want to try testing the DNS resolver in with an integration test to remove the additional indirection between InstanceConnectionNamerResolver -> DNS Resolver -> DNS resolution. If you tested with a real DNS record, you could remove the indrection.

I'll leave this @jackwotherspoon to sign-off on.

core/src/main/java/com/google/cloud/sql/core/DnsResolver.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import java.util.Collection;
+import javax.naming.NameNotFoundException;
+
+interface DnsResolver {
+  Collection<String> resolveTxt(String domainName) throws NameNotFoundException;
```
jackwotherspoon, line 23: we are using `dnsName` in other places, is `domainName` the same thing or is it a different meaning? If they are one and the same we should consolidate usage to one naming pattern

hessjcg: Go connector uses `DomainName` as the configuration field. 

I considered `fqdn`, `dnsName`, `domainName`, `host`, and `hostname`. FQDN or FullyQualifiedDomainName is too wordy, and dnsName is an expands to "domain name system name", which is self-redundant. I was tempted `host`, or `hostname` because that is commonly what it is called as part of the a URL [RFC 1738](https://datatracker.ietf.org/doc/html/rfc1738#section-3.1). I decided against it because I wanted to be specific that it needs to contain the domain name. 

Go connector uses `DNSInstanceConnectionNameResolver` for the resolver class name, and `DNSResolver` for the singleton instance. This refers to the fact that it uses the DNS system to resolve the `DomainName` into an instance connection name. In retrospect, this felt too wordy, so the class name is `DNSResolver` in java.

core/src/main/java/com/google/cloud/sql/core/JndiDnsResolver.java
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * @param dnsServer the DNS server hostname
+   * @param port the DNS server port (DNS servers usually use port 53)
+   */
+  JndiDnsResolver(String dnsServer, int port) {
+    this.jndiPrefix = "dns://" + dnsServer + ":" + port + "/";
+  }
+
+  /**
+   * Returns DNS records for a domain name, sorted alphabetically.
```
sanmahapatra, line 47: For a domain name, we should only get 1 TXT record if one exists. Why do we have Collection<String> for the return type?

hessjcg: We expect exactly 1 DNS record for an instance. However, it is possible that the connector could receive 0, 1, or many DNS records. Other application logic will gracefully handle cases where the number of responses is not 1 as expected. This class is only responsible for resolving DNS and reporting the results. Thus its data model is in line with the DNS protocol, returning a collection that may contain 0 or more responses.

sanmahapatra (approved):
LGTM

PR #2048, opened 2024-07-29
chore: turn off test log noise by default

enocom:
If developers wish to re-enable debug logs, change the root level to "debug."


jackwotherspoon (approved):
LGTM, think you just need a header for the file

PR #2051, opened 2024-08-13
chore: Update the google-github-actions/auth configuration to use secrets

hessjcg:
This the google-github-actions/auth configuration was incorrectly changed to use `vars` in #2046 
to comply with security team requests. However, this should use github secrets, which hold the service_account and workload_identity_provider. The secret vars are not yet set when the google-github-actions/auth task is run.


jackwotherspoon:
Closing this based on discussion offline

PR #2056, opened 2024-08-15
feat: Automatically reset DNS-configured connections on DNS change. Part of #2043

### LINKED ISSUE #2043: Automatically configure and fail-over connections using DNS

hessjcg:
Poll for changes to the DNS entry every 30 seconds and update the connector to connect to the new
instance when the value of the DNS TXT record changes. Close all existing sockets connected to the old 
instance. 

See also: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/844

Fixes #2043


jackwotherspoon (approved):
LGTM 👍

PR #2060, opened 2024-08-26
feat: Support CAS server certificate authority to validate instance connections.

hessjcg:
For Cloud SQL instances with CAS enabled, the connector will use the certificate chain of trust reported by the 
SQL Admin API to validate the instance connection.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/Connector.java
+++ b/core/src/main/java/com/google/cloud/sql/core/Connector.java
@@ -118,6 +118,7 @@ class Connector {
       SSLSocket socket = (SSLSocket) metadata.getSslContext().getSocketFactory().createSocket();
       socket.setKeepAlive(true);
       socket.setTcpNoDelay(true);
+
       socket.connect(new InetSocketAddress(instanceIp, serverProxyPort));
 
       try {
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
@@ -45,9 +45,9 @@ import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -84,10 +84,15 @@ class DefaultConnectionInfoRepository implements ConnectionInfoRepository {
   }
 
   // Creates a Certificate object from a provided string.
-  private Certificate createCertificate(String cert) throws CertificateException {
+  private List<Certificate> parseCertificateChain(String cert) throws CertificateException {
     byte[] certBytes = cert.getBytes(StandardCharsets.UTF_8);
     ByteArrayInputStream certStream = new ByteArrayInputStream(certBytes);
-    return CertificateFactory.getInstance("X.509").generateCertificate(certStream);
+    List<Certificate> certificates = new ArrayList<>();
+    while (certStream.available() > 0) {
+      Certificate c = CertificateFactory.getInstance("X.509").generateCertificate(certStream);
+      certificates.add(c);
+    }
+    return certificates;
   }
 
   private String generatePublicKeyCert(KeyPair keyPair) {
@@ -296,18 +301,17 @@ class DefaultConnectionInfoRepository implements ConnectionInfoRepository {
                     + "IP address.",
                 instanceName.getConnectionName()));
       }
-
       // Update the Server CA certificate used to create the SSL connection with the instance.
       try {
-        Certificate instanceCaCertificate =
-            createCertificate(instanceMetadata.getServerCaCert().getCert());
+        List<Certificate> instanceCaCertificates =
+            parseCertificateChain(instanceMetadata.getServerCaCert().getCert());
 
         logger.debug(String.format("[%s] METADATA DONE", instanceName));
 
         return new InstanceMetadata(
             instanceName,
             ipAddrs,
-            Collections.singletonList(instanceCaCertificate),
+            instanceCaCertificates,
             "GOOGLE_MANAGED_CAS_CA".equals(instanceMetadata.getServerCaMode()),
             instanceMetadata.getDnsName(),
             pscEnabled);
@@ -371,7 +375,9 @@ class DefaultConnectionInfoRepository implements ConnectionInfoRepository {
     // Parse the certificate from the response.
     Certificate ephemeralCertificate;
     try {
-      ephemeralCertificate = createCertificate(response.getEphemeralCert().getCert());
+      // The response contains a single certificate. This uses the parseCertificateChain method
+      // to parse the response, and then uses the first, and only, certificate.
+      ephemeralCertificate = parseCertificateChain(response.getEphemeralCert().getCert()).get(0);
     } catch (CertificateException ex) {
       throw new RuntimeException(
           String.format(
@@ -407,8 +413,7 @@ class DefaultConnectionInfoRepository implements ConnectionInfoRepository {
           KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
       kmf.init(authKeyStore, new char[0]);
 
-      TrustManagerFactory tmf =
-          InstanceCheckingTrustManagerFactory.newInstance(instanceName, instanceMetadata);
+      TrustManagerFactory tmf = InstanceCheckingTrustManagerFactory.newInstance(instanceMetadata);
 
       SSLContext sslContext;
 
@@ -428,7 +433,6 @@ class DefaultConnectionInfoRepository implements ConnectionInfoRepository {
           sslContext = SSLContext.getInstance("TLSv1.2");
         }
       }
-
       sslContext.init(kmf.getKeyManagers(), tmf.getTrustManagers(), new SecureRandom());
 
       logger.debug(
--- a/core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManagerFactory.java
+++ b/core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManagerFactory.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
 import javax.net.ssl.TrustManagerFactory;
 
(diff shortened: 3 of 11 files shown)


core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -281,15 +287,24 @@ private InstanceMetadata fetchMetadata(CloudSqlInstanceName instanceName, AuthTy
...
-
       // Update the Server CA certificate used to create the SSL connection with the instance.
       try {
-        Certificate instanceCaCertificate =
+        // TODO:
+        // if("GOOGLE_MANAGED_CAS_CA".equals(instanceMetadata.getServerCaMode())) {
+        //
+        // } else {
+        //
+        // }
```
jackwotherspoon, line 297: any reason why this is a TODO?

core/src/main/java/com/google/cloud/sql/core/InstanceMetadata.java
```diff
@@ -18,24 +18,48 @@
...
+
+  InstanceMetadata(
+      Map<IpType, String> ipAddrs,
+      List<Certificate> instanceCaCertificates,
+      boolean casManagedCertificate,
+      String dnsName) {
+    this.ipAddrs = ipAddrs;
+    this.instanceCaCertificates = instanceCaCertificates;
+    this.casManagedCertificate = casManagedCertificate;
+    this.dnsName = dnsName;
```
jackwotherspoon, line 47: do we really need two constructors for `InstanceMetadata` ? seems like we should be able to just use one...

jackwotherspoon (changes requested):
I believe we want to update the PSC check now that dnsName can be for CAS-based instances as well as PSC just like Go did: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/850/files#diff-5097a430f11cfd2c79c874367e2f96cf048d9e47112b01f7650d961d8a677ff8R104

Not sure where exactly in the Java code base this would be, maybe here? https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/e13fa46e110bea16b7e4088687c073f1ee615388/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java#L282

hessjcg:
Java has this requirement covered without needing additional code: 

> I believe we want to update the PSC check now that dnsName can be for CAS-based instances as well as PSC just like Go did: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/850/files#diff-5097a430f11cfd2c79c874367e2f96cf048d9e47112b01f7650d961d8a677ff8R104

The Java connector calls SSLSocket.connect() at Connector.java:118 with either the instance IP or the PSC domain name. When the instance is a CAS instance, the TLS context will be configured to use the built-in logic to check certificate trust. When SSLSocket.connect() is called with the PSC domain name, the built-in TLS logic will check the certificate's SAN records as expected.

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManagerFactory.java
```diff
@@ -43,18 +44,27 @@
...
+    // If this is using a CAS-managed certificate, don't add the InstanceCheckingTrustManagerFactory
+    // to check the CN against the instance name.
+    if (instanceMetadata.isCasManagedCertificate()) {
+      delegate.init(trustedKeyStore);
+      return delegate;
+    }
+
+    // Otherwise, Use a custom trust manager factory that checks the CN against the instance name
     InstanceCheckingTrustManagerFactory tmf =
         new InstanceCheckingTrustManagerFactory(instanceName, delegate);
```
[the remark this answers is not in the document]
jackwotherspoon: +1 we need to actually check the dnsName against the SAN fields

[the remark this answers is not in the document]
hessjcg: Fixed. I also split some of the code from this into 2 other PRs: 

1. #2067 
2. (this one) feat: Support CAS server certificate authority to validate instance connections
3. #2066

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManger.java
```diff
@@ -92,6 +95,62 @@ private void checkCertificateChain(X509Certificate[] chain) throws CertificateEx
...
+          "Instance metadata for " + instanceMetadata.getInstanceName() + " has an empty dnsName");
+    }
+    for (String san : sans) {
+      if (san.equalsIgnoreCase(dns)) {
+        return;
+      }
+    }
+    throw new CertificateException(
+        "Server certificate does not contain expected name '"
+            + instanceMetadata.getInstanceName()
```
feng-zhe, line 119: Use `instanceMetadata.getDnsName()` instead.

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManger.java
```diff
@@ -92,6 +95,62 @@ private void checkCertificateChain(X509Certificate[] chain) throws CertificateEx
...
+    ArrayList<String> names = new ArrayList<>();
+
+    Collection<List<?>> sanAsn1Field = cert.getSubjectAlternativeNames();
+    if (sanAsn1Field == null) {
+      return names;
+    }
+
+    for (List item : sanAsn1Field) {
+      Integer type = (Integer) item.get(0);
+      if (type == 0 || type == 2) {
```
feng-zhe, line 134: Can we add a comment here for the 0 and 2? I assume that `0` is the `otherName` and `2` is the `dNSName` according to the RFC 5280.

feng-zhe (approved):

core/src/main/java/com/google/cloud/sql/core/Connector.java
```diff
@@ -118,6 +119,10 @@ Socket connect(ConnectionConfig config, long timeoutMs) throws IOException {
       SSLSocket socket = (SSLSocket) metadata.getSslContext().getSocketFactory().createSocket();
       socket.setKeepAlive(true);
       socket.setTcpNoDelay(true);
+
+      SSLParameters sslParams = new SSLParameters();
+      socket.setSSLParameters(sslParams);
```
jackwotherspoon, line 124: can you add a comment in the code here explaining this new piece?

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -367,7 +374,7 @@ private Certificate fetchEphemeralCertificate(
     // Parse the certificate from the response.
     Certificate ephemeralCertificate;
     try {
-      ephemeralCertificate = createCertificate(response.getEphemeralCert().getCert());
+      ephemeralCertificate = createCertificate(response.getEphemeralCert().getCert()).get(0);
```
jackwotherspoon, line 377: Let's add a comment in the code about which cert `.get(0)` is gettiing...

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -424,9 +430,14 @@ private SslData createSslData(
           sslContext = SSLContext.getInstance("TLSv1.2");
         }
       }
-
       sslContext.init(kmf.getKeyManagers(), tmf.getTrustManagers(), new SecureRandom());
 
+      // If this is CAS-managed, then use the HTTPS algorithm to check the subjectAlternateNames
+      // against the certificate.
```
jackwotherspoon, line 436: should we not also be doing this for PSC? It also has the DNS name as a SAN entry...

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManagerFactory.java
```diff
@@ -39,32 +40,38 @@
  * <p>class ConscryptWorkaroundTrustManager - the workaround for the Conscrypt bug.
  *
  * <p>class InstanceCheckingTrustManager - delegates TLS checks to the default provider and then
- * checks that the Subject CN field contains the Cloud SQL instance ID.
+ * does custom hostname checking in accordance with these rules:
+ *
+ * <p>If an instanceMetadata.casEnabled Check that the subjectAlternativeNames in the server
```
jackwotherspoon, line 45: nit: this sentence seems off to me, reads a bit funny...

jackwotherspoon (changes requested):

core/src/main/java/com/google/cloud/sql/core/Connector.java
hessjcg: Removed. This was unnecessary.

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
hessjcg: Done.

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
hessjcg: Actually, this is wrong. I'm removing it. It won't work because unlike Go we cannot override the hostname passed to Socket.connect(). 

I built custom hostname verification logic in `InstanceCheckingTrustManager`. 

The built-in HTTPS hostname verification algorithm won't work because the connector cannot override the hostname when `Socket.connect(addr)` is called with an IP Address.

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManagerFactory.java
hessjcg: Good catch. I fixed it.

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManger.java
hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManger.java
hessjcg: You are correct. I added a comment to clarify.

core/src/main/java/com/google/cloud/sql/core/InstanceCheckingTrustManger.java
```diff
@@ -92,6 +95,66 @@ private void checkCertificateChain(X509Certificate[] chain) throws CertificateEx
...
+      if (san.equalsIgnoreCase(dns)) {
+        return;
+      }
+    }
+    throw new CertificateException(
+        "Server certificate does not contain expected name '"
+            + instanceMetadata.getDnsName()
+            + "' for Cloud SQL instance "
+            + instanceMetadata.getInstanceName());
+  }
```
jackwotherspoon, line 122: is this not just the regular SSL flow? Is there benefit to this custom logic?

jackwotherspoon (approved):
LGTM ✅

PR #2067, opened 2024-09-13
chore(refactor): Add CAS related fields to the InstanceMetadata.

hessjcg:
Refactor InstanceMetadata to include fields needed to validate the TLS server certificates from CAS-
enabled instances. This contains only changes to the data structure of InstanceMetadata. It does not change
to the certificate validation logic.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
@@ -47,6 +47,7 @@ import java.security.cert.X509Certificate;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -277,8 +278,13 @@ class DefaultConnectionInfoRepository implements ConnectionInfoRepository {
         }
       }
 
-      // resolve DnsName into IP address for PSC
-      if (instanceMetadata.getDnsName() != null && !instanceMetadata.getDnsName().isEmpty()) {
+      // If PSC is enabled, resolve DnsName into IP address for PSC
+      boolean pscEnabled =
+          instanceMetadata.getPscEnabled() != null
+              && instanceMetadata.getPscEnabled().booleanValue();
+      if (pscEnabled
+          && instanceMetadata.getDnsName() != null
+          && !instanceMetadata.getDnsName().isEmpty()) {
         ipAddrs.put(IpType.PSC, instanceMetadata.getDnsName());
       }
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/google/cloud/sql/core/InstanceMetadata.java
```diff
@@ -18,24 +18,55 @@
...
+      Map<IpType, String> ipAddrs,
+      List<Certificate> instanceCaCertificates,
+      boolean casManagedCertificate,
+      String dnsName,
+      boolean pscEnabled) {
+    this.instanceName = instanceName;
     this.ipAddrs = ipAddrs;
-    this.instanceCaCertificate = instanceCaCertificate;
+    this.instanceCaCertificates = instanceCaCertificates;
+    this.casManagedCertificate = casManagedCertificate;
```
feng-zhe, line 44: nit: may be rename this to `isCasManagedCertificate`.

feng-zhe (approved):

core/src/main/java/com/google/cloud/sql/core/InstanceMetadata.java
hessjcg: Usually we don't prefix private boolean fields with "is" this in java. The accessor method is prefixed with "is". 

```java 
class InstanceMetadata {
private boolean casManagedCertificate
public boolean isCasManagedCertificate() {
  return casManagedCertificate;
}
}
```

feng-zhe (approved):

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -298,7 +299,14 @@ private InstanceMetadata fetchMetadata(CloudSqlInstanceName instanceName, AuthTy
...
 
-        return new InstanceMetadata(ipAddrs, instanceCaCertificate);
+        return new InstanceMetadata(
+            instanceName,
+            ipAddrs,
+            Collections.singletonList(instanceCaCertificate),
+            "GOOGLE_MANAGED_CAS_CA".equals(instanceMetadata.getServerCaMode()),
+            instanceMetadata.getDnsName(),
+            instanceMetadata.getPscEnabled() != null
+                && instanceMetadata.getPscEnabled().booleanValue());
```
jackwotherspoon, line 309: Could we not use the fact that if the `ipAddrs` map contains an `IPType.PSC` entry then PSC is enabled...?

hessjcg: Sure, but I want to keep instanceMetadata as a very simple value without any logic. So I'd rather compute pscEnabled here. 

I updated also updated logic on line 282 to only add the PSC address to the ipAddrs map if pscEnabled == true

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -298,7 +304,13 @@ private InstanceMetadata fetchMetadata(CloudSqlInstanceName instanceName, AuthTy
...
         logger.debug(String.format("[%s] METADATA DONE", instanceName));
 
-        return new InstanceMetadata(ipAddrs, instanceCaCertificate);
+        return new InstanceMetadata(
+            instanceName,
+            ipAddrs,
+            Collections.singletonList(instanceCaCertificate),
+            "GOOGLE_MANAGED_CAS_CA".equals(instanceMetadata.getServerCaMode()),
+            instanceMetadata.getDnsName(),
+            pscEnabled);
```
jackwotherspoon, line 313: did we decide to have this here? I thought we were going to use the presence of the PSC ip type in the map to check PSC enabled?

jackwotherspoon (approved):
LGTM

PR #2080, opened 2024-10-23
docs: Add missing javadoc comments to public methods.

hessjcg:
The checkstyle rules are becoming more strict. All public methods must now have javadoc.


jackwotherspoon (approved):
LGTM, can we add a quick description to the PR explaining why we are adding this for future ref?

PR #2095, opened 2024-12-17
feat: Support Customer CAS Private CA for server certificates.

hessjcg:
Support instances configured with a Customer managed CAS private CA. This includes the integration test.


core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -312,7 +312,7 @@ private InstanceMetadata fetchMetadata(CloudSqlInstanceName instanceName, AuthTy
             instanceName,
             ipAddrs,
             instanceCaCertificates,
-            "GOOGLE_MANAGED_CAS_CA".equals(instanceMetadata.getServerCaMode()),
+            isCasManagedCertificate(instanceMetadata),
```
jackwotherspoon, line 315: Would it simplify things to have this function take in `serverCaMode` instead of `instanceMetadata`?

Would make adding a standalone unit test for `isCasManagedCertificate` easy and reduce having to call `.getServerCaMode()` twice in the function.

```suggestion
            isCasManagedCertificate(instanceMetadata.getServerCaMode()),
```

jackwotherspoon (changes requested):

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
hessjcg: I've inverted the logic here. Going forward, both GOOGLE_MANAGED_CAS_CA, CUSTOMER_MANAGED_CAS_CA, and future new kinds of CA we will do standard TLS domain name validation using the server certificate SAN records. The certificate validation logic for the original `GOOGLE_MANAGED_INTERNAL_CA` is now the exception. The code should reflect it: 

We don't have a unit test for this method, but this change is covered by this test: `com.google.cloud.sql.core.ConnectorTest#create_successfulPublicCasConnection`

jackwotherspoon: I still think it would be simpler for the function to take in `serverCaMode` instead of `instanceMetadata`...

hessjcg: Fixed.

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepository.java
```diff
@@ -332,6 +332,18 @@ private InstanceMetadata fetchMetadata(CloudSqlInstanceName instanceName, AuthTy
...
+   * Instances with serverCaMode == GOOGLE_MANAGED_INTERNAL_CA or serverCaMode == null use a legacy,
+   * non-standard server certificate validation strategy. All other cases serverCaMode use a
+   * standard TLS hostname validation using the SubjectAlternativeNames records.
+   *
+   * @param serverCaMode from the instance metadata.
+   * @return true when the instance uses a CAS certificate, and should use standard validation.
+   */
+  private static boolean isCasManagedCertificate(String serverCaMode) {
+    return serverCaMode != null && !"GOOGLE_MANAGED_INTERNAL_CA".equals(serverCaMode);
+  }
```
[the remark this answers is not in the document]
hessjcg: Good catch. Updated again for clarity.

jackwotherspoon (approved):
LGTM ✅

PR #2097, opened 2025-01-08
test: Integration test for Customer Private CA CAS instance.

hessjcg:
Adds an integration test to ensure the java connector works with a Customer Private CA instance.


hessjcg:
This integration test is included in #2095

PR #2098, opened 2025-01-09
chore: Add domain name to the connector config.

hessjcg:
Add DomainName to connection config, and an InstanceNameResolver function. This will allow the connector
to be configured using a domain name referencing the instead of the instance name.

This is part of the implementation of #2043. 

See #2043 for the whole feature definition.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstanceName.java
+++ b/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstanceName.java
@@ -31,10 +31,42 @@ class CloudSqlInstanceName {
   // Some legacy project ids are domain-scoped (e.g. "example.com:PROJECT:REGION:INSTANCE")
   private static final Pattern CONNECTION_NAME =
       Pattern.compile("([^:]+(:[^:]+)?):([^:]+):([^:]+)");
+
+  /**
+   * The domain name pattern in accordance with RFC 1035, RFC 1123 and RFC 2181.
+   *
+   * <p>Explanation of the Regex:
+   *
+   * <p>^: Matches the beginning of the string.<br>
+   * (?=.{1,255}$): Positive lookahead assertion to ensure the domain name is between 1 and 255
+   * characters long.<br>
+   * (?!-): Negative lookahead assertion to prevent hyphens at the beginning.<br>
+   * [A-Za-z0-9-]+: Matches one or more alphanumeric characters or hyphens.<br>
+   * (\\.[A-Za-z0-9-]+)*: Matches zero or more occurrences of a period followed by one or more
+   * alphanumeric characters or hyphens (for subdomains).<br>
+   * \\.: Matches a period before the TLD.<br>
+   * [A-Za-z]{2,}: Matches two or more letters for the TLD.<br>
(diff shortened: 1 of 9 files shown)


jackwotherspoon (commented):
This LGTM but let's also get someone more experienced at Java to review it as well

pandirigoog:
Test comment!

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstanceName.java
```diff
@@ -35,14 +35,26 @@ class CloudSqlInstanceName {
...
+
+  /**
+   * Initializes a new CloudSqlInstanceName class containing the domain name.
+   *
+   * @param connectionName instance connection name in the format "PROJECT_ID:REGION_ID:INSTANCE_ID"
+   * @param domainName the domain name used to look up the instance, or null.
+   */
+  CloudSqlInstanceName(String connectionName, String domainName) {
     this.connectionName = connectionName;
+    this.domainName = domainName;
```
pandirigoog, line 57: Do we want to validate it with some DOMAIN_NAME matcher ? generic DNS name or something ?

core/src/main/java/com/google/cloud/sql/core/Connector.java
```diff
@@ -154,6 +157,37 @@ ConnectionInfoCache getConnection(ConnectionConfig config) {
     return instance;
   }
 
+  private ConnectionConfig resolveConnectionName(ConnectionConfig config) {
```
pandirigoog, line 160: I know its private method (so method documentation is optional) but I think this method is important enough that it requires some explanation. Like cloudSqlInstance is prioritized over domainName when both are set etc.

core/src/test/java/com/google/cloud/sql/core/ConnectorTest.java
```diff
@@ -131,6 +131,50 @@ public void create_successfulPrivateConnection() throws IOException, Interrupted
     assertThat(readLine(socket)).isEqualTo(SERVER_MESSAGE);
   }
 
+  @Test
+  public void create_successfulPublicConnectionWithDomainName()
```
pandirigoog, line 135: Can we have a test for both domainName and cloudSqlInstance set?

core/src/main/java/com/google/cloud/sql/core/CloudSqlInstanceName.java
hessjcg: Done. I added a Regex to validate the domain name.

core/src/main/java/com/google/cloud/sql/core/Connector.java
hessjcg: Done.

core/src/test/java/com/google/cloud/sql/core/ConnectorTest.java
hessjcg: Done.

pandirigoog (approved):

jackwotherspoon (approved):

pandirigoog (approved):

PR #2102, opened 2025-01-13
docs: Add docs for connector configuration using DNS Names, Part of #2043

hessjcg:
Adds documentation to describe how to use the hostname for JDBC and R2DBC.

### DIFF
--- a/docs/jdbc.md
+++ b/docs/jdbc.md
@@ -563,6 +563,123 @@ Properties connProps = new Properties();
 connProps.setProperty("cloudSqlRefreshStrategy", "lazy");
 ```
 
+
+
+### Using DNS domain names to identify instances
+
+The connector can be configured to use DNS to look up an instance. This would
+allow you to configure your application to connect to a database instance, and
+centrally configure which instance in your DNS zone.
+
+#### Configure your DNS Records
+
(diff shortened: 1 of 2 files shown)


docs/jdbc.md
```diff
@@ -563,6 +563,123 @@ Properties connProps = new Properties();
...
+
+**Note:** You are strongly discouraged from adding DNS records for your
+Cloud SQL instances to a public DNS server. This would allow anyone on the
+internet to discover the Cloud SQL instance name.
+
+For example: suppose you wanted to use the domain name
+`prod-db.mycompany.example.com` to connect to your database instance
+`my-project:region:my-instance`. You would create the following DNS record:
+
+- Record type: `TXT`
```
pandirigoog, line 587: Should we also document about A record types? Like the mapping from DNS name rto IPs?

hessjcg: I think that's better left for the official Google Cloud Help docs. The connectors don't use the A record. Maybe we add a link once the feature is released.

pandirigoog: Sounds good.

jackwotherspoon (approved):

PR #2103, opened 2025-01-13
chore: Use domain name from R2DBC connections, Part of #2043.

hessjcg:
Now, the connector will detect if the R2DBC `HOST` connection property contains a domain name or a 
Cloud SQL instance name. If `HOST` has a domain name, the connector will look up the domain name to
determine the instance name. 

For example, if the R2DBC url is set to `r2dbc:gcp:mssql://<DB_USER>:<DB_PASS>@db.example.com/db`,
the connector will lookup the TXT record for `db.example.com`. The record should contain a valid
Cloud SQL instance name. 

Part of #2043

### DIFF
--- a/r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
+++ b/r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
@@ -124,11 +124,22 @@ public abstract class GcpConnectionFactoryProvider implements ConnectionFactoryP
             ? RefreshStrategy.LAZY
             : RefreshStrategy.BACKGROUND;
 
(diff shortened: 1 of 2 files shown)


r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -124,11 +124,22 @@ public ConnectionFactory create(ConnectionFactoryOptions connectionFactoryOption
             ? RefreshStrategy.LAZY
             : RefreshStrategy.BACKGROUND;
 
+    final String r2dbcHostname = (String) connectionFactoryOptions.getRequiredValue(HOST);
```
pandirigoog, line 127: Can we add a test for this?

pandirigoog (approved):

jackwotherspoon (approved):

PR #2107, opened 2025-01-23
refactor: Remove unnecessary pre-fetching from R2DBC.

hessjcg:
When the java connector was refactored to use ConnectorConfig and ConnectionConfig value objects, it was no longer necessary to pre-fetch the connection configuration to ensure that enableIamAuth was set correctly. Also, this is inconsistent behavior compared to the JDBC socket factory.

### DIFF
--- a/r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
+++ b/r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
@@ -144,9 +144,6 @@ public abstract class GcpConnectionFactoryProvider implements ConnectionFactoryP
                     .withRefreshStrategy(refreshStrategy)
                     .build())
             .build();
-    // Precompute SSL Data to trigger the initial refresh to happen immediately,
-    // and ensure enableIAMAuth is set correctly.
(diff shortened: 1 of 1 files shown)


r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java
```diff
@@ -146,7 +146,7 @@ public ConnectionFactory create(ConnectionFactoryOptions connectionFactoryOption
             .build();
     // Precompute SSL Data to trigger the initial refresh to happen immediately,
     // and ensure enableIAMAuth is set correctly.
-    InternalConnectorRegistry.getInstance().getConnectionMetadata(config);
+    // InternalConnectorRegistry.getInstance().getConnectionMetadata(config);
```
jackwotherspoon, line 149: can we just remove these three lines instead of commenting it out?

hessjcg: Yes. Sorry, that fix went into the wrong PR.

jackwotherspoon (approved):

PR #2123, opened 2025-03-05
fix: Add GraalVM configuration for the JDK's DNS Resolver.

hessjcg:
This fixes the GraalVM distribution so that it can resolve DNS Names.


jackwotherspoon (approved):
LGTM ✅

PR #2125, opened 2025-03-12
feat: Use standard TLS hostname validation for instances with DNS names.

hessjcg:
When the connector is configured with a DNS name, or if the Cloud SQL Instance reports that it has a DNS Name,
the connector will use standard TLS hostname validation when checking the server certificate. Now, the server's
TLS certificate must contain a SAN record with the instance's DNS name.

The ConnectSettings API added a field dns_names which contains all of the valid DNS names for
an instance..

See also https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/954


jackwotherspoon (approved):
LGTM, as long as we have manually tested on a real instance 👍

PR #2137, opened 2025-04-02
test: update IP type env variable on integration tests

kgala2:
Configures integration tests to use the `IP_TYPE` environment variable for connector testing. This variable, set during build setup (defaulting to PUBLIC), determines the IP type used in tests (`ipTypes`).


hessjcg (approved):
Nice work!

PR #2161, opened 2025-05-20
chore: Superficial annotations for Deprecated and test runner.

hessjcg:
This contains superficial changes to annotations. This tags specific fields labeled as deprecated in the javadoc
with `@Deprecated` annotation. Also, this adds `@RunWith(JUnit4.class)` to several test cases to make 
the test classes explicit.


kgala2 (approved):
LGTM

PR #2193, opened 2025-08-17
feat: Add DnsJavaResolver and related test

panavenue:
Description:

This PR Adds the following:

1. The dnsjava dependency, and a new class `DnsJavaResolver`. Which will be use to replace `JndiDnsResolver` for DNS record lookup. 
2. DnsJavaResolverTest to make sure DnsJavaResolver are working as expected

Next:
Will replace and remove `JndiDnsResolver` with `DnsJavaResolver` in the next PR.

### DIFF
--- a/core/src/main/java/com/google/cloud/sql/core/DnsJavaResolver.java
+++ b/core/src/main/java/com/google/cloud/sql/core/DnsJavaResolver.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.sql.core;
+
+import java.net.UnknownHostException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.stream.Collectors;
+import javax.naming.NameNotFoundException;
+import org.xbill.DNS.Lookup;
+import org.xbill.DNS.Record;
+import org.xbill.DNS.SimpleResolver;
+import org.xbill.DNS.TXTRecord;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/cloud/sql/core/DnsJavaResolver.java
```diff
@@ -0,0 +1,109 @@
...
+      // 3. Execute the DNS query.
+      lookup.run();
+
+      // 4. Check the result of the lookup.
+      int resultCode = lookup.getResult();
+      if (resultCode == Lookup.HOST_NOT_FOUND) {
+        throw new NameNotFoundException("DNS record not found for " + domainName);
+      }
+      if (resultCode == Lookup.TYPE_NOT_FOUND) {
+        return Collections.emptyList();
```
hessjcg, line 84: This should also throw a NameNotFoundException.

core/src/test/java/com/google/cloud/sql/core/DnsJavaResolverTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    // Check that the record contains the expected IP address.
+    assertThat(records).contains(VALID_DOMAIN_NAME_DATA);
+  }
+
+  @Test
+  public void testResolveTxt_notFound() throws NameNotFoundException {
+    DnsJavaResolver resolver = new DnsJavaResolver();
+    Collection<String> records = resolver.resolveTxt(INVALID_DOMAIN_NAME);
+    assertThat(records).isEmpty();
```
hessjcg, line 47: This should return name not found exception.

core/src/main/java/com/google/cloud/sql/core/DnsJavaResolver.java
```diff
@@ -0,0 +1,109 @@
...
+      }
+
+      // 5. Process the records, sort them, and return.
+      Record[] records = lookup.getAnswers();
+      if (records == null || records.length == 0) {
+        return Collections.emptyList();
+      }
+
+      // A single TXT record can contain multiple strings, so we use flatMap.
+      return Arrays.stream(records)
```
hessjcg, line 98: Add tests to make sure this logic works correctly. You may want to mock `Lookup` to do this.

hessjcg (commented):
Nice work.  A few small comments.

core/src/main/java/com/google/cloud/sql/core/DnsJavaResolver.java
panavenue: Instead of mocking, I am wondering if I should create another TXT dns record in the project (One that contains multiple strings). So this is testing everything e2e?

hessjcg (approved):

PR #2200, opened 2025-09-09
feat: Metadata Exchange support for the Java connector.

hessjcg:
This is the client-side implementation of the metadata exchange protocol for Cloud SQL instances. This will
allow connectors to pass metadata about the connection to the Cloud SQL instance while the connection
is being established.

In particular, this is a part of resolving the issues with the Auth Proxy connecting to MySQL instances that
use the caching_sha2_password extension. (See https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/2317)

This does not update the default MDX behavior yet. That will be added in #2207


core/src/main/java/com/google/cloud/sql/core/ConnectionConfig.java
```diff
@@ -244,7 +254,22 @@ public ConnectionConfig withDomainName(String domainName) {
         authType,
         unixSocketPathSuffix,
         domainName,
-        connectorConfig);
+        connectorConfig,
+        mdxClientProtocolType);
+  }
+
+  /** Creates a new instance of the ConnectionConfig with an updated cloudSqlInstance. */
```
panavenue, line 261: `cloudSqlInstance` -> `clientProtocolType`?

core/src/main/java/com/google/cloud/sql/core/DefaultConnectionInfoRepositoryFactory.java
```diff
@@ -48,6 +48,11 @@ public DefaultConnectionInfoRepository create(
     return new DefaultConnectionInfoRepository(adminApiBuilder);
   }
 
+  @Override
+  public String getUserAgent() {
```
panavenue, line 52:
```suggestion
  public String getUserAgents() {
```

hessjcg: Fixed.

.kokoro/check_coverage.sh
```diff
@@ -16,7 +16,20 @@
 # `-e` enables the script to automatically fail when a command fails
 set -e
 
-export CUR_COVER=$(cat core/target/site/jacoco/index.html | grep -o 'Total[^%]*' | sed 's/<.*>//; s/Total//')
+#export CUR_COVER=$(cat core/target/site/jacoco/index.html | grep -o 'Total[^%]*' | sed 's/<.*>//; s/Total//')
```
panavenue, line 19: nit: should we remove this line?

.kokoro/check_coverage.sh
```diff
@@ -16,7 +16,20 @@
...
+# Calculate total coverage, skipping the generated protobuf file.
+# The CSV file generated by Jacoco has the test coverage for each. We use AWK to process this CSV file.
+# here's what the AWK program means:
+#
+#     NR > 1 &&                                   Skip the first line, it only has field names
+#       $3 != "com.google.cloud.sql.core.mdx"     Skip any line for class in the mdx package
+#     { covered += $6; total += $6+$5; }          For each line add a running sum of covered lines and total lines
+#     END { printf "%2.f", covered/total * 100}'  At the end, print out the coverage percentage, no decimals
+#
+CUR_COVER=$( awk -F,  \
```
panavenue, line 30: I see line #19 has `export` to make this var available to sub process. I wonder if `export` is needed for this var?

panavenue (approved):
LGTM with some nits

panavenue (dismissed):

core/src/main/java/com/google/cloud/sql/core/ConnectionConfig.java
```diff
@@ -244,7 +254,22 @@ public ConnectionConfig withDomainName(String domainName) {
         authType,
         unixSocketPathSuffix,
         domainName,
-        connectorConfig);
+        connectorConfig,
+        mdxClientProtocolType);
+  }
+
+  /** Creates a new instance of the ConnectionConfig with an updated clientProtocolType. */
+  public ConnectionConfig withMdxClientProtocolType(String clientProtocolType) {
```
kgala2, line 262: nit: should the parameter say mdxClientProtocolType?

kgala2 (approved):

core/src/main/java/com/google/cloud/sql/core/ConnectionConfig.java
hessjcg: Fixed.

PR #2229, opened 2025-11-22
docs: Add Cloud Run and JDBC connector examples

kgala2:


hessjcg (commented):
Let's add a Cloud Run + Spring Boot + MySQL example also. Lots of our users like Spring Boot. I don't think we need to replicate it to all the other database types. Create it in the `examples/cloudrun/springboot-mysql` folder.

PR #2276, opened 2026-05-04
fix: Update dependencies to latest as of 2026-05-04

hessjcg:
Update dependencies to latest.


pom.xml
```diff
@@ -842,6 +824,44 @@
       </build>
   </profile>
 
+  <profile>
```
hessjcg, line 827: Errorprone only supports JDK 21+ now. Only run it for newer JDK versions. Skip it for older JDK versions.

kgala2 (approved):

